PR #6, opened 2016-06-21
2.1: Note about automatic populating of the self_update parameter in the Kerner Options

hustodemon:
This is for 2.1 - is this the correct branch?

It should go to doc for 3.0 as well, but it seems the file with corresponding contents doesn't exist (was looking in the develop branch).


hustodemon:
@keichwa Could you please take a look at this PR when you get a minute?

keichwa:
Thanks.  Sorry for the delay!

PR #11, opened 2016-09-22
Be more clear on the hostname update

schaefi:
Try to make the hostname change more clear

### DIFF
--- a/xml/public-cloud-setup.xml
+++ b/xml/public-cloud-setup.xml
@@ -129,22 +129,22 @@
        also required to adapt the <filename>/etc/hosts</filename>
        file as follows:
        <screen>$ vi /etc/hosts
-$ echo "${ip_address} suma.cloud.net suma" >> /etc/hosts</screen>
-       The current value for the <replaceable>ip_address</replaceable>, can
-       be obtained from the public cloud Web console or from within a terminal
-       session as follows:
+$ echo "${internal_address} suma.cloud.net suma" >> /etc/hosts</screen>
+       The current value for the <replaceable>internal_address</replaceable>,
+       can be obtained from the public cloud Web console or from within
+       a terminal session as follows:
        <itemizedlist>
        <listitem>
         <para> Obtain public ip address from within Amazon EC2 instance </para>
-        <screen>$ ec2metadata --public-ipv4</screen>
+        <screen>$ ec2metadata --local-ipv4</screen>
        </listitem>
        <listitem>
         <para> Obtain public ip address from within Google Compute Engine instance </para>
-        <screen>$ gcemetadata --query instance --network-interfaces --external-ip</screen>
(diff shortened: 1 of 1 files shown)


schaefi:
For reading compiled here:

https://w3.suse.de/~ms/Project/doc-susemanager/build/susemanager-cloud/single-html/susemanager-cloud/

xml/public-cloud-setup.xml
```diff
@@ -110,14 +110,26 @@
       <para> Set the hostname to a name of your choice. Please note
        it is important to provide the <replaceable>system name</replaceable>
        and not the <replaceable>fully qualified host name</replaceable>
-       to the command. In the following example we use
-       <emphasis role="strong">suma</emphasis> as hostname and the
-       domain name is set to <emphasis role="strong">cloud.net</emphasis>.
-       </para>
-      <screen>$ hostnamectl set-hostname suma
-$ hostname -f
+       to the hostnamectl command.
```
aosthof: I'd put hostnamectl in <filename> tags to be consistent.

xml/public-cloud-setup.xml
```diff
@@ -110,14 +110,26 @@
...
+
+       On the assumption that the fully qualified host name of the
```
aosthof: As I'm not sure if 'On the assumption that ...' is a common English phrase (Robert can tell for sure), I'd rather write 'Assuming that ..."

xml/public-cloud-setup.xml
```diff
@@ -110,14 +110,26 @@
...
+       instance created in the cloud service is initially set to
+       <emphasis role="strong">cloud_instance.cloud.net</emphasis>,
+       this would result in <emphasis role="strong">cloud_instance</emphasis>
+       as the machine name and <emphasis role="strong">cloud.net</emphasis>
```
aosthof: Here you write 'machine name', but above you called it 'system name'. Replace the one or the other to be consistent.

xml/public-cloud-setup.xml
```diff
@@ -110,14 +110,26 @@
...
+       defining the domain name.
 
-suma.cloud.net</screen>
+       In the following example we will change the machine name to
```
aosthof: Same here - machine name vs. system name

schaefi: Thanks I've updated the inconsistencies

xml/public-cloud-setup.xml
rjschwei: We should say:

It is expected that the Fully Qualified Domain Name (FQDN) is set by the cloud framework, in this example we use cloud_instance.cloud.net

rjschwei:
OK, just the wording change w.r.t. "assume" the rest looks good to me.

schaefi:
Thanks I've update the docs. However I'm confused by Joel's last test report.

@keichwa: Please merge this once Robert or Alex gives us a thumbs up, thanks

PR #13, opened 2016-10-20
Fix the epub build

hustodemon:
...by adding missing `role` attribute.


jcayouette:
This feels like a feature that daps should support. Why should a user enter two separate rolls for outputs? Daps should support this simply by supplying the output then adjust image formats automatically.

keichwa:
Yeah, this is not that easy ;)  AFAIK, most things are handled in the style sheets (xsl and css), but additionally and traditionally we must provide two blobs--one for fo (that's PDF) and one for html (that's HTML and epub).  Theoretically, you can also provide one for txt output...

Of course, it would be nice to make this less error prone.  The less keying in XML, the better.

PR #43, opened 2016-12-14
BP: onboarding rate not a problem anymore

moio:
Recent Manager builds can onboard hundreds of minions at once without problems. This section can be removed.


jcayouette:
Thanks +1

PR #44, opened 2017-01-16
Explain the database connection formula

moio:
This is to explain better the database connection formula, as I believe the parameters in it were not really explained elsewhere.

@jcayouette can you take a look at the DocBook formatting? :smile:


jcayouette:
Looks good thank you! Nice work with the block elements. screen and itemized should always remain outside the para tag. Currently we don't need to specify the language option as our suse.com/doc scripts from Novell are not using my  syntax highlighting work https://highlightjs.org/

PR #53, opened 2017-10-19
remove client OSses which are out of maintenance

kkaempf:


jcayouette:
Thanks!

kkaempf:
https://www.suse.com/lifecycle is the official resource

We might want to add a "best effort" notice about unsupported OSses

kkaempf:
Something like

"About versions out-of-mainenance

We acknowledge the need to manage older versions or service pack level of SLES or RHEL in order to bring them to a supported level with the help of SUSE Manager. SUSE Manager engineering will try to provide 'best effort' support in case of problems. Please talk to your SUSE consultant or sales representative."

jcayouette:
Once the adoc effort is complete we will get a note added.

PR #61, opened 2018-03-08
Ke smt rmt

keichwa:


xml/bp_chap_choosing_dist_scheme.xml
```diff
@@ -18,63 +18,173 @@
...
+    <step>
+     <para>
+      Export settings about repositories to mirror to the specified path
+      (in this case, <filename>/mnt/usb</filename>); this command will
+      create a <filename>repos.json</filename> file there:
+     </para>
+     <screen>&prompt.root;rmt-cli export settings /mnt/usb</screen>
+     <para>
+      Check <filename>repos.json</filename> and edit it to remove the
+      repositories you do not need.
```
ikapelyukhin: @keichwa While in theory it is possible to edit the JSON file, the documentation **absolutely must not** suggest this to the end-user. :scream: 

The user should choose which repositories they want to mirror using `rmt-cli repos enable` and `rmt-cli products enable` commands (i.e., `rmt-cli product enable sles/15/x86_64` to enable SLES15 repos). After enabling the repos, they will be written to `repos.json` file (which is used to transfer data from online RMT to offline RMT instance -- SUMA doesn't use this file) and the repos configured in this file will be exported into the export directory.

The user can also verify which repos are currently enabled with `rmt-cli repos list` command.

xml/bp_chap_choosing_dist_scheme.xml
```diff
@@ -128,14 +238,23 @@ smt-mirror --dbreplfile /tmp/dbrepl.xml --directory /media/disk \
...
+  </sect2>
+  
+  <sect2 xml:id="disconnect.mgr.update-repos">
+   <title>Updating Repositories on &susemgr; From Storage Media</title>
+   
   <para>
-   The next procedure will show you how to update your server from the SMT
-   media.
+   <remark>Will this work the same way with media created on
+   RMT?</remark> This procedure will show you how to update the
```
ikapelyukhin: We hope so :grimacing: At least, that's the intention.

PR #70, opened 2018-05-07
Ke reference systemdetails virtualization

keichwa:


keichwa:
minimal commit to address https://github.com/SUSE/spacewalk/issues/4256

PR #83, opened 2018-05-18
Virtual Host Manager: clarify that manual creation of the file is not supported

moio:
We previously had slightly contradictory content:

"If you are using an unsupported virtual host manager, you can create and use a file formatted in JSON containing information..."

Then, in the highlighted "note" box:

"The file-based is not meant to be used with manually crafted files. It only meant to be used with the output of virtual-host-gatherer against some other module."

https://bugzilla.suse.com/show_bug.cgi?id=1047253#c26


Bischoff:
Thanks, it's already much better.

I'm not sure that providing a sample of the JSON file makes sense if no one is supposed to touch it... But I can understand it is left if we want to explain a bit what's behind.

moio:
Intention is also to reassure the anxious VMWare responsible that the SUSE Manager responsible isn't exfiltrating any private data in the process.

PR #96, opened 2018-06-04
Lower requirements for virt guests management.

Bischoff:
In the case the user just wants to restart VMs, stop them, etc,
it is not needed to register them.


Bischoff:
Thanks Karl. I don't know if a backport is needed.

keichwa:
E Bischoff <notifications@github.com> writes:

> Thanks Karl. I don't know if a backport is needed.

Thus far, nobody asked about it.  Assuming no, though.  These days
back-ports are cumbersome because we additionally must convert from adoc
to xml...

-- 
                           90409 Nürnberg, Germany

PR #98, opened 2018-06-06
minor edits to syntax

Loquacity:


keichwa:
Your edit looks nice, Lana!

Tag usage is still a little bit in flux, but I now use menu: for real navigation stuff only (ideally, that's menu:Main Menu[Systems] or menu:Main Menu[Systems > Systems]).  Then we have this cool btn:[xxx]. And for all the rest of the GUI labels I use this ugly [guimenu]``text text`` construct.

To be discussed ;)

Loquacity:
Yes, it seems a bit uneven at the moment. I'll keep going! Thanks :)

Loquacity:
What do you think of using the backticks instead?

keichwa:

> What do you think of using the backticks instead?

``label`` instead of [guimenu]``label``?  This would result in "lighter"
output (HTML and PDF).  We first convert to XML (DocBook) and then to
the final formats.  I think it works this way:

``label``    ==>  <emphasis>label</emphasis>  ==>  just italics

[guimenu]    ==>  <guimenu>label</guimenu>    ==>  something greenish


Ir depends on what we want to accomplish...

-- 
                           90409 Nürnberg, Germany

jcayouette:
Both are fine for me, and consistency would be the goal here :)
As we already have a stylesheet rule for guimenu, I would suggest sticking with [guimenu]``label`` for now. I can go ahead and add it to our Asciidoctor GS if you both are ok with this.

Loquacity:
Yep, that sounds good to me, thanks for the clarification :)

PR #107, opened 2018-06-22
Issue #4985 Adding to troubleshooting section from wiki pages

Loquacity:

PR #111, opened 2018-06-27
update version string

Loquacity:
Fixing comment from https://github.com/SUSE/doc-susemanager/pull/107


keichwa:
Wondering whether we can use the {sles-version} macro in the [path] context?  Then, for SP, we probably must add a lowercase variant in entities.adoc:

:sp-version-l: sp3

kkaempf:
@keichwa good point. We should have something like 'latest major release' (12 today, 15 in a few weeks), and 'latest service pack' (SP3 today, GA in a few weeks).
cc @jcayouette

Loquacity:
That seems to have worked quite well :)

PR #113, opened 2018-06-27
Issue #498: Update required ports table

Loquacity:
* Updated with info from https://wiki.microfocus.com/index.php/SUSE_Manager/Security_Firewall
* Made wording parallel, and added info about traffic type where missing
* Added column for protocol (TCP/UDP)


keichwa:
according to the GS, salt makes use of port 80.

Loquacity:
According to https://wiki.microfocus.com/index.php/SUSE_Manager/Ports_Diagrams port 80 isn't used much at all, so I've removed it. But now I think about it, having it there with a note is probably better. Let me know what you think :)

adoc/bp_chap_intro.adoc
```diff
@@ -192,17 +192,19 @@ The following tables provide you with an overview of ports which are used by {su
...
-| 4505 | Required by the Salt-master to accept communication requests via TCP from minions. Inbound.
-| 4506 | Required by the Salt-master to accept communication requests via TCP from minions. Inbound.
+| Port | Protocol | Description
+| 22   | TCP      | SSH
+| 67   | TCP/UDP  | DHCP
+| 69   | TCP/UDP  | TFTP, used to support PXE services
+| 80   | TCP      | HTTP, used in some bootstrap cases
+| 123  | UDP      | NTP time service
+| 443  | TCP      | HTTPS, used for {webui}, client, and proxy server traffic
+| 4505 | TCP      | Salt, used by the Salt-master to accept communication requests from minions
```
moio: Salt uses two ZeroMQ sockets between the master and the minion (or proxy). Each of those consume one TCP connection.

First is on port 4505, connection is initiated by the minion and it is always open. master will send commands to minions through this channel (called a publish/subscribe topology, as the minion subscribes to notifications from the master).

Second is on port 4506, connection is initiated by the minion and it is kept open only when needed. In particular, minions will open this port when they have to report results to the master (typically, but not necessarily, after a command that had been received on port 4505 has finished). This is called a request/response topology as the minion sends "requests" to the master.

https://docs.saltstack.com/en/getstarted/system/communication.html

I think the above is good information to have somewhere here, or linked from here.

adoc/bp_chap_intro.adoc
```diff
@@ -192,17 +192,19 @@ The following tables provide you with an overview of ports which are used by {su
...
+| 4506 | TCP      | Salt, used by the Salt-master to accept communication requests from minions
+| 5222 | TCP      | XMPP client, used for communications with the [systemitem]``osad`` daemon on client systems
```
moio: Please specify traditional clients. This is not used with minions

moio:
@jcayouette didn't we have such a list somewhere else in the manuals? My memory gets flaky :smiley:

Loquacity:
@moio AFAICT, we have it in BP_chap_intro, advanced_topics_ports, and quickstart3_chap_install_overview (which is a copy of the table in BP). Are there other locations i've missed?

Loquacity:
@keichwa @moio OK to merge this?

PR #115, opened 2018-06-28
Updated migration instructions from 3.1 to 3.2

Loquacity:
Migration instructions in the BP Migration chapter only had migration from 3.0 to 3.1. Updated to reflect migration from 3.1 to 3.2. Reported by customer at https://bugzilla.suse.com/show_bug.cgi?id=1099348

### DIFF
--- a/adoc/bp_sp_migration.adoc
+++ b/adoc/bp_sp_migration.adoc
@@ -52,13 +52,9 @@ endif::[]
 [[bp.sp.migration.sp.intro]]
 == Service Pack Migration Introduction
 
-You can upgrade the underlying operating system and also migrate {susemgr} server from one patch level to the other (SP migration) or from one version to the next.
-This works for migrating {susemgr} server 3.0 to version 3.1, or version 3.1 to version {productnumber}.
-For migrating version 3.0 to version 3.1, see the product documentation coming with {susemgr} 3.1:
-
-----
-https://www.suse.com/documentation/suse-manager-3/
-----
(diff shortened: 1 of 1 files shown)


keichwa:
I guess we also still want to mention how you migrate from 3.0 to 3.2.  And we must check whether 3.2 on SLES 12 SP2 (or earlier) is possible and supported.  Often @mcalmer can help with this version information.

kkaempf:
> I guess we also still want to mention how you migrate from 3.0 to 3.2.

Actually, not at this point. We might get such requests in the future, but currently skipping 3.1 is not supported.

3.2 on SP2 is also not supported and never will be.

cc @keichwa

keichwa:
It's still confusing/confused and kind of half done ;)  Sorry, this is always easy in theory, but not that easy to cover all aspects.

SLES 12 SP2 will never be the base for SUMA 3.2 (see Klaus' comment above).  This means customers must ensure to migrate SLES 12 to SP3 before or during the SUMA migration.  We probably still want to keep postgres warning.  I guess yast/zypper will offer working or supported scenarios only.

We will also need fresh yast screen shot(s) and zypper output listings...

kkaempf:
Release notes (update released last week) clearly state now (as prerequisites for 3.1 to start the upgrade to 3.2):

**SUSE Manager 3.1 Server must be on the latest patch level (3.1.6 as of this writing), based on SLES 12 SP3, and have PostgreSQL 9.6 running.**

PR #116, opened 2018-07-02
Issue #498: new AT bare metal provisioning section

Loquacity:
From https://wiki.microfocus.com/index.php/SUSE_Manager/NewFeatures2.1/BareMetalProvisioning

### DIFF
--- a/adoc/advanced_topics_cobbler.adoc
+++ b/adoc/advanced_topics_cobbler.adoc
@@ -83,7 +83,7 @@ To PXE boot systems, you will require a DHCP server, or have access to a network
 .Correct Hostname Configuration
 ====
 Cobbler uses hostnames as a unique key for each system.
-If you are using the [option]``pxe-default-image`` to onboard baremetal systems, make sure every system has a unique hostname.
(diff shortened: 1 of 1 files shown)


Loquacity:
Note that, after some research, I've gone with "bare metal" over "bare-metal" or "baremetal". If there's an established precedent for a different option, though, I'm happy to change it.

kkaempf:
:+1:

keichwa:
That's probably fine.  in the documentation team, I usually delegate those decisions to @sknorr .

Wondering whether we want to keep the term "bare metal" at all.  ATM, the term "unprovisioned system" is also in use...

sknorr:
Thanks for the hint, @keichwa ... opened https://github.com/SUSE/doc-styleguide/issues/80

Mmh, question though: isn't "bare-metal" just a jargony term for "physical system" (i.e. not a virtualized system)? Or is this used differently in Suma?

kkaempf:
@sknorr yes, "bare metal" and "physical" is almost the same. In SUMA, "bare metal" is mostly used to denote "unprovisioned" (no OS installed) systems, while "physical" refers to "OS running non-virtualized".

(This gets tricky on zSeries or Power systems :wink: )

PR #117, opened 2018-07-02
Issue #5086: Remove entities in heading

Loquacity:
https://github.com/SUSE/spacewalk/issues/5086

This removes the entities and cleans up the heading until we can take a closer look at fixing them.


kkaempf:
Hmm, we have entities for a reason and should keep them. Esp. in the light of Uyuni.

Loquacity:
i was just doing this as a fix until we know the source of the problem

PR #118, opened 2018-07-02
Issue #4998 Added info about reclaiming disk space

Loquacity:
Also some general tidying up in this section


Loquacity:
Oh, it probably should be {productname}. I'll wait for @jcayouette to comment.

PR #119, opened 2018-07-02
Issue #5086 remove entities from headings

Loquacity:
https://github.com/SUSE/spacewalk/issues/5086

It turns out that adoc can't render the entiities in headings at all, so I've reverted them to verbatim text.

### DIFF
--- a/adoc/advanced_topics_suma3_zsystems.adoc
+++ b/adoc/advanced_topics_suma3_zsystems.adoc
@@ -1,5 +1,5 @@
 [[art.suma.install.zsystems]]
-= {susemgr} on IBM {zseries}
+= SUSE Manager on IBM Z Series Systems
 ifdef::env-github,backend-html5,backend-docbook5[]
 //Admonitions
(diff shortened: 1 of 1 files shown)


kkaempf:
Then let's fix our tooling. I don't want verbatim.

Loquacity:
Klaus: I agree we need to fix the tooling (or at least put some rules in place about how to handle this kind of situation), but I really just want to get the current public documentation fixed for our customers as soon as possible. Is it OK if we use verbatim for the moment, as a workaround to fix the public docs, and then determine the best way forward?

kkaempf:
Agreed, lets use verbatim for the moment and

a) open an issue for our tooling (which tooling did you use ? what are the steps to reproduce this problem ?)
b) keep this issue open so we don't forget to replace the verbatim with entities later

Loquacity:
Closing this, as I seem to have it working locally with entities. I will need to investigate further to work out what on earth happened here, but for now I can create an updated copy to send live.

PR #131, opened 2018-07-19
Updated overview for readability

Loquacity:
Structure changes, wording updates, etc


Loquacity:
Yeah, so every instance of {productname} at the moment is rendering blank, which is why I've changed them back. That is on my list of things I need to talk to you about ;)

jcayouette:
These are rendering blank because we need to add the attribute at the top of each document, such as:
```
:productname: keep_Value_Blank
```
 Then we will use the correct make build command. The command replaces the entity depending on if you build for Uyuni or Manager.

PR #135, opened 2018-07-23
RES 7 has a different filesystem structure

keichwa:
fix spacewalk-repo-sync command - https://github.com/SUSE/spacewalk/issues/5226


Loquacity:
Changes look good, but we should probably implement a RES entity. I bet we're contravening RH branding by using the acronym.

keichwa:
ok, thank.  I'll merge this and then create a second PR concerning source refinements (there is more than just the entity issue in this file).  It would be better to do it the other way around...

PR #136, opened 2018-07-24
Updated susemgr entity throughout

Loquacity:
_DO NOT MERGE!_

@jcayouette this is still rendering blank for me on a local build. Can you work out what I've done wrong?


jcayouette:
Merged after fixing makefile, I will test and revert if we hit bugs.

PR #141, opened 2018-07-26
Forbid dot characters in the configuration channels label

hustodemon:
These are not allowed anymore because they cause problem in salt (salt interprets `.` as a separator character).

@jcayouette This change should go to 3.2 documentation. I created a PR against `develop` - is that correct or should I have used another branch?


hustodemon:
Update: @jcayouette This needs to go to 3.1 as well. I'll create a PR against `maintenance/3.1`, is that procedure correct? :smile:

PR #145, opened 2018-07-30
[WIP] SUSE Manager for Retail reference documentation

aaannz:
This is still in WIP, but feel free to add commits.

I am not sure if this will be distributed as a standalone book or part of the SUMA documentation.

Missing parts:

- [ ]  Saltboot process
- [ ] Image customization
- [ ] Retail specific settings for generic formulas (DNC, TFTP, PXE, VSFTP)

### DIFF
--- a/adoc/reference-retail.adoc
+++ b/adoc/reference-retail.adoc
@@ -23,34 +23,25 @@ It allows administrators to configure network topologies of selected servers, bu
 2. layer consists of one or more servers providing local network services and boot server services. These servers are generally referred to as branch servers
 3. layer consists of deployed machines e.g. point of sale machines. These machines are generally referred to as terminals
 
+=== Generic installation workflow
+
+This is a checklist of tasks to prepare generic {smr} stack:
+
+1. install and configure central {susemgr} server
+2. install {smr} related pattern? formula? TODO: PM,PrjM
+3. install and configure SLES12SP3 to be used as a build host
+- register build host to {susemgr} using salt
+- set build host as a OS Image build host in {susemgr} web user interface and apply highstate
+4. install and configure branch server
+- register branch server to {susemgr} using salt
+5. assign and configure branch server formulas
+
 === Generic Workflow
 Having installed and configured {smr} infrastructure, generic workflow is as follows:
 
 1. build image to be deployed on terminals
-+
-Standard OS Image building feature of {susemgr} is used. Images are built from customized Kiwi templates, as a starting point see https://github.com/SUSE/manager-build-profiles/tree/master/OSImage[SUSE provided OS Image templates].
-+
-Adapt these templates to contain all required software, branding, configurations. Refer to https://doc.opensuse.org/projects/kiwi/doc/[Kiwi  Image System documentation] for details how to work with Kiwi templates.
-+
-If you are creating Kiwi template from scratch, please make sure `image/preferences/type` elements attribute `boot` is set to `saltboot/suse-SLES12`
-+
-Once the image is build, it will be automatically transferred to {susemgr} server
-
 2. synchronize images from {susemgr} server to branch servers
-+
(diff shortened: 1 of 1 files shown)


jcayouette:
I have a few thoughts I think we should address/clarify before merging,

- Reference refers to the SUSE Manager WebUI, In contrast this should probably be renamed to "SUSE Manager for Retail" or some other similar format.
- Instead of the {smr} entity I propose we stick with {productname} and simply generate docs at build time, with a separate included section for retail. This means you would have the exact same data "no duplication" of SUSE Manager docs, and the maintained "retail" doc with simply an entity change.
- Let's find out how we may best distribute this,  how does the retail PO feel about the above idea? After these have been discussed then we can merge it. :)

Currently we can maintain a Makefile containing our build commands. We could just following this logic for simplicities sake:

make suma - :productname: SUSE Manager
make uyuni - :productname: Uyuni
make retail - :productname: SUSE Manager for Retail

The retail doc would not be included in Uyuni/suma built documentation by a conditional statement.

P.S. I realize last week there was a mailing list discussion but I was unable to follow as effectively as I would have liked due to HA priority. Ill try and catch up as time permits.

Loquacity:
This is a great start for content, thank you @aaannz! <3

Before we get too far down this road, though, we need to sort a few things out:
[ ] The title and structure of this book
[ ] How we are going to handle entities (in this doc, and in the rest of the doc suite). 
[ ] Whether or not we are publishing a Retail-specific documentation suite (in addition to this doc), or just pointing Retail readers to the SUMA docs. 

It's super-important that we get the planning done *before* we start writing content. I am going to create a document to allow us to get started on that, and will add a link here shortly.

aaannz:
Thanks for all the comments. I share all of you concerns.

@jcayouette:
- what reference to Web UI you have in mind? Retail is part of the SUMA, everything is being distributed together and there is no special branding AFAIK. We are basically creating a documentation for a specific use case of SUMA.
- I'll change `{smr}` to `{productname}`
- Distribution question is for @jsrain or @abhinav-puri: Should the Retail documentation be standalone or part of the SUMA?

@Loquacity:
- Indeed, I started it as a reference guide, but this is probably something bigger. Still, it is only a one special use case to SUMA. I started with content immediately so that QA guys can have some starting point for their work.

keichwa:
Ondřej Holeček <notifications@github.com> writes:

> - I'll change `{smr}` to `{productname}`

I'd recommend to stick with `{smr}`, at least for moment.  It's easy to
change it later, if needed.

> - Indeed, I started it as a reference guide, but this is probably
> something bigger.

I see your point, but ATM we use the term "reference" resp. "Reference
Guide" for the document that describes the SUMA Web UI, one by one menu
item.  Thus I'd like us to call the document about the Retail product
that you started, just "SUMA for Retail" or "SUMA for Retail Guide", or
something similar.

> Still, it is only a one special use case to SUMA. I started with
> content immediately so that QA guys can have some starting point for
> their work.

Yes, I also appreciate this!  We can re-write part of it or restructure
it later, if we know what we want.  If you have free cycles, please
continue with drafting a "SUMA for Retail Guide" :-)

-- 
                           90409 Nürnberg, Germany

Loquacity:
> Indeed, I started it as a reference guide, but this is probably something bigger. Still, it is only a one special use case to SUMA. I started with content immediately so that QA guys can have some starting point for their work.

That's great, I'm sure they appreciate it :)

Once we've got a plan in place (https://github.com/SUSE/doc-susemanager/wiki/SUSE-Manager-for-Retail-Documentation/88847b8cd661e1b3ae96ad889a0f31903e0fa82e) we can create a new doc and put all this straight in there. Thanks for all the hard work <3

Loquacity:
This is now being handled here: https://github.com/SUSE/doc-susemanager/pull/172

PR #154, opened 2018-08-03
Fix: specify which AK can be used when image building

mbologna:
Add a tiny note about activation keys and image building.

### DIFF
--- a/adoc/advanced_topics_image_management.adoc
+++ b/adoc/advanced_topics_image_management.adoc
@@ -121,6 +121,8 @@ salt '$your_minion' state.highstate
 
 Create an activation key associated with the channel that your images will use.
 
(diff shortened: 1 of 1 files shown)


adoc/advanced_topics_image_management.adoc
```diff
@@ -121,6 +121,8 @@ salt '$your_minion' state.highstate
 
 Create an activation key associated with the channel that your images will use.
 
+NOTE: only activation keys associated with a channel that is not "SUSE Manager Default" can be used to build Containers.
```
Loquacity: NOTE: To build containers, you will need an activation key that is associated with a channel other than "SUSE Manager Default".

Loquacity: Additionally, this should probably be in an admonition. Let me know if you want me to do the changes.

PR #162, opened 2018-08-09
Switch clients from traditional to salt

Loquacity:
https://github.com/SUSE/spacewalk/issues/5335

And some editing of the first para in the section.

### DIFF
--- a/adoc/quickstart3_chap_suma_keys_and_first_client.adoc
+++ b/adoc/quickstart3_chap_suma_keys_and_first_client.adoc
@@ -350,7 +350,7 @@ This section describes the first method and uses a bootstrap repository.
 The second method uses bootstrap script, and is described in <<registering.clients.traditional>>.
 The third method uses the {webui}, and is described in <<ref.webui.systems.bootstrapping>>.
 
-You can also use these methods to change existing traditional clients into Salt clients.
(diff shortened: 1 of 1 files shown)


keichwa:
We are having fun with adoc markup (as expected)...  Some remarks:

You seem to prefer [btn]``xxx`` over btn:[xxx].

A package name is not a command ;)

Wondering whether we want to stick with the rule using menu:xxx only for entries of the left navigation menu and, maybe, cascaded tabs on sysmte details pages, etc.?  For widgets inside of a page, I usually [guimenu]``checkbox_label``.

But all this is not that important to me, thus I'll approve your PR!

Loquacity:
Hrm, there's some weirdness here, I'm sure these are changes you already did. Anyway, I've reverted them.

On a related note, though, partly I think the problem is that adoc doesn't have the range of docbook. In docbook, you can be really specific about the thing you are marking up (right down to <systemitem class="xx">!), but adoc just doesn't have that ability, so I'm kind of using [path] for everything that doesn't have a specific semantic. That said, I just discovered [systemitem] so it's my new favourite catch all.

The [btn] thing is just me forgetting that it works differently to the others (I also have to look up the menu one every time still) ;)

PR #166, opened 2018-08-13
Proxies can also be minions now (FATE#323069)

dincamihai:
Update documentation to reflect changes made as part of this card https://github.com/SUSE/spacewalk/issues/5303


dincamihai:
@mantel do you think we need to add something else here?

keichwa:
I noticed that the chapter requires more changes.  We must check all procedure whether they equally reflect traditional and Salt clients.  (then there are other unrelated issues with the chapter such as pseudo links or broken markup intro'ed during xml2adoc conversion...

I guess I should open new issues...

PR #171, opened 2018-08-21
Update Makefile

keichwa:
Use PRODUCTNAME setting when preparing xml targets
better sorting of targets (suma comes first)


keichwa:
received approval via irc/mail

PR #175, opened 2018-08-29
virtualiazation documentation update

cbosdo:
This PR documents uyuni-project/uyuni#5
An update of image ``system_details_traditional_virtualization_details.png``

### DIFF
--- a/adoc/advanced_topics_virtualization.adoc
+++ b/adoc/advanced_topics_virtualization.adoc
@@ -96,12 +96,11 @@ To automatically register {vmguest}s with {productname}, a bootstrap script is n
 ==== Activation Keys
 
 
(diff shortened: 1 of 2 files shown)


keichwa:
Thanks for the update and special thanks for attacking the xml2adoc conversation artifacts!  If nobody takes action, I can probably review and merge it next week.  Great work, you did not put in obstacles, but there are still some bits from the conversation that I'd like to eliminate.  BTW, we also have ``[package]`` as a tag for package names.

cbosdo:
@keichwa Thanks for the info. Just note that the [resource] use was here before: I just changed the formatting of the text. But I can change it too if you need it

keichwa:
Cederic, I hope you can agree with the adoc related changes I committed.

cbosdo:
@keichwa looks good to me so far, no real content change FWICT

PR #192, opened 2018-09-18
Revert "do not install 3rd party software"

keichwa:
Reverts SUSE/doc-susemanager#191


Loquacity:
Why the revert?

keichwa:
> Why the revert?

I did not want to merge directly, but first waiting for your review.

adoc/quickstart3_chap_suma_setup_with_yast.adoc
```diff
@@ -17,6 +17,14 @@ You will perform the following procedures:
 [[suma-setup-with-yast-sumasetup]]
 == {productname} Setup
 
+[WARNING]
+.Third Party Software Not Allowed
+--
+{productname} is a extension of {sls} and thus compatible the with software shipped with {sls}.
```
Loquacity: > {productname} is an extension of {sls}, so it is compatible with all {sls} default software.

(Or similar. There's a couple of typos in there, plus I really dislike 'thus' :P )

mcalmer:
@keichwa - btw. It is not true that we do not allow any 3rd Party Software. AFAIK monitoring software is allowed as long as they do not exchange any basic libraries like ssl, crypto software and similar things.

mcalmer:
@keichwa - I added Klaus as reviewer as he need to verify if the statement is correct.

kkaempf:
Thanks for updating this @keichwa. There's one caveat to be added (please adapt wording as you see fit): "in case of problems, SUSE reserves the right to ask to remove any 3rd party software (and associated configuration changes) and to reproduce the problem on a 'pristine' system."

Loquacity:
Wording is fine, but I have some reservations about use of the word "pristine" (especially with the markup). I think "clean" is a more widely used (and understood) term, and allow sus to remove the italics. 

Ref: https://books.google.com/ngrams/graph?content=clean+system%2Cpristine+system&year_start=1970&year_end=2018&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cclean%20system%3B%2Cc0%3B.t1%3B%2Cpristine%20system%3B%2Cc0

keichwa:
Yes, "clean" is probably better understood and from the surrounding text it should be clear what is meant.

OTOH, in the free software/opensource world 'pristine' has a very special meaning: untouched sources (without patches or changes) or a non-customized system and those things.  But you are right we must avoid those slang words.

kkaempf:
Please also push it to the 3.2 branch (for the upcoming 3.2.3 update) !

PR #212, opened 2018-09-26
fix entity defs

keichwa:
In our context, smt != rmt (on version 12 you can continue using smt).
The others are probably obvious.

We probably also want it in maint/3.2.

### DIFF
--- a/adoc/entities.adoc
+++ b/adoc/entities.adoc
@@ -44,7 +44,7 @@
 :openscap: OpenSCAP
 :uarr:
 :ncc: Novell Customer Center
-:smtool: Subscription Management
+:smtool: Subscription Management Tool
 :rmtool: Repository Management Tool
 :vmguest: VM Guest
 :vmhost: VM Host Server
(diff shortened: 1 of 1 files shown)


adoc/entities.adoc
```diff
@@ -26,15 +27,15 @@
...
+:slea: SLE
 :sleda: SLED
 :ay: AutoYaST
 :baseos:
 :deploy: Deployment Guide
 :upgrade: Upgrade Guide
 :yast: YaST
 :rootuser: root
-:mdash: -
+:mdash: —
```
Loquacity: :grin:

adoc/entities.adoc
```diff
@@ -43,8 +44,7 @@
 :openscap: OpenSCAP
 :uarr:
 :ncc: Novell Customer Center
-:sp-ver: {sp-version}
-:smtool: {rmtool}
+:smtool: Subscription Management
```
Loquacity: Should this be "Subscription Management Tool"?

keichwa: > Should this be "Subscription Management Tool"?

;)

PR #225, opened 2018-10-02
WIP: backport from develop

keichwa:
more or less cosmetic


Loquacity:
Why are you editing the XML and not the adoc?

keichwa:
> Why are you editing the XML and not the adoc?

I think 3.1 is XML-only.  That one of the reasons why we do not like back-porting to 3.1 that much ;)

Loquacity:
Ah, that makes sense :100:

PR #227, opened 2018-10-03
Updated wording around >5000 terminals, per Abhinav's feedback.

Loquacity:


aaannz:
I talked with @moio and I share his view that we should not print any number here. In my PR #229 I removed this paragraph completely.

Loquacity:
OK, I've had a long chat with @moio and discovered that no such restriction exists. Therefore, I'm removing the statement from this section, and adding a general statement to the final chapter about contacting consulting for assistance. Hopefully this satisfies everyone :)

PR #229, opened 2018-10-03
Retail documentation update

aaannz:
This PR fixes
* hardware requirements - previous were completely unrealistic, now they are in line with SUMA requirements
* remove need for `spacewalk.api` configuration - it is no longer required
* remove limits about terminal numbers
* SUMA configuration part contained some info about formula installation which is not relevant to SMR. Replaced by need to sync SUSE channels


aaannz:
Adapted PR with requested changes.

aaannz:
Rebased and fixed one procedure headings. What do you mean by sentence case headings?

PR #231, opened 2018-10-08
Ke qs31 bootstrap

keichwa:


keichwa:
I hope that fulfills the def of done.  I think we should not spend that much time on this issue for version 3.1.

PR #232, opened 2018-10-09
Fixed table layout

Loquacity:
According to https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#tables you need the 'a' signifier in the cols attribute to say that there's asciidoc content in the column. From there, you need to make sure that there's a line above and below each list.


keichwa:
Thanks for your help.  I'm a table hater.  Most of the time they are superfluous and used for (bad) layout purposes.  Here table usage is okayish, but repeating "supported" again and again is tedious.  I think we better just use some kind of check marker and say somewhere what is means.

PR #233, opened 2018-10-09
Technical review feedback

Loquacity:
From @Bischoff

### DIFF
--- a/adoc/retail_chap_install.adoc
+++ b/adoc/retail_chap_install.adoc
@@ -96,6 +96,7 @@ Each procedure is detailed in this section.
 === Procedure Install the Central Server
 
 For instructions to install {susemgr}, see the xref:quickstart3_chap_install_overview.adoc[{susemgr} Getting Started Guide].
+
 [WARNING]
 .Do Not Enable PXE Boot
 ====
@@ -148,9 +149,10 @@ zypper in -t pattern suma_retail
 * dhcpd-formula
 * image-sync-formula
 * pxe-formula
+* saltboot-formula
 * susemanager-retail-tools
 * tftp-formula
-* vsftpd-formula
+* vsftpd-formula (optional)
 +
 Install any missing packages with [command]``zypper``:
 +
--- a/adoc/retail_chap_install.adoc
+++ b/adoc/retail_chap_install.adoc
@@ -152,7 +152,7 @@ zypper in -t pattern suma_retail
 * saltboot-formula
 * susemanager-retail-tools
 * tftp-formula
-* vsftpd-formula (optional)
+* vsftpd-formula
 +
 Install any missing packages with [command]``zypper``:
 +
--- a/adoc/retail_chap_introduction.adoc
+++ b/adoc/retail_chap_introduction.adoc
@@ -10,12 +10,12 @@ include::entities.adoc[]
 {smr} {productnumber} is an open source infrastructure management solution, optimized and tailored specifically for the retail industry.
 It uses the same technology as {susemgr}, but is customized to address the needs of retail organizations.
 
-It is designed for use in retail situations where customers can use point-of-service terminals to purchase or exchange goods, take part in promotions, or collect loyalty points.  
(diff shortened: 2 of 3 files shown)


adoc/retail_chap_introduction.adoc
```diff
@@ -4,29 +4,29 @@
...
---
+[[retail.sect.intro.what]]
+== What is {smr}?
 
 {smr} {productnumber} is an open source infrastructure management solution, optimized and tailored specifically for the retail industry.
+It uses the same technology as {susemgr}, but is customized to address the needs of retail organizations.
 
-This document provides an overview of {smr}, and guides you through initial installation and setup.
-It should be read in conjunction with the {susemgr} documentation suite, available from https://www.suse.com/documentation/suse-manager-3/
+It is designed for use in retail situations where customers can use point-of-service terminals to purchase or exchange goods, take part in promotions, or collect loyalty points.
```
Bischoff: Maybe just add "e.g." or "for example" or "among others" to state that this list is not exhaustive.

aaannz: I think this sentence is not needed. One of the usages can very well be school computers, or self-service kiosks in banks, hospitals, etc. (as was in old SLEPOS). I don't really like limiting our use cases again and again just and only for retail, despite the name.

Bischoff: Then add these other use cases.

And to avoid limiting, say "for example" or "among others".

I feel this sentence is very much needed.

adoc/retail_chap_introduction.adoc
```diff
@@ -4,29 +4,29 @@
...
-This document provides an overview of {smr}, and guides you through initial installation and setup.
-It should be read in conjunction with the {susemgr} documentation suite, available from https://www.suse.com/documentation/suse-manager-3/
+It is designed for use in retail situations where customers can use point-of-service terminals to purchase or exchange goods, take part in promotions, or collect loyalty points.  
 
-For more information about managing your {smr} environment, or to find out where to get help, see <<retail_chap_next.adoc#retail.chap.next>>.
---
+{smr} is intended for use in installations that include servers, workstations, point-of-service terminals, and other devices.
+It allows administrators to manage, configure, and update the software on their servers, and manage the deployment and provisioning of point-of-service machines.
+
+{smr} is designed to help you reduce costs, optimize operations, and ensure compliance while increasing the flexibility, reliability, and uptime of your point-of-service devices.
```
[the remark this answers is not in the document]
Bischoff: yes, last paragraph could be removed.

adoc/retail_chap_introduction.adoc
Loquacity: Agree, I think it's a good representation of use cases which is otherwise lacking in the docs (and always the number one thing readers ask for!). I've added a para to elaborate.

adoc/retail_chap_install.adoc
```diff
@@ -148,9 +149,10 @@ zypper in -t pattern suma_retail
 * dhcpd-formula
 * image-sync-formula
 * pxe-formula
+* saltboot-formula
 * susemanager-retail-tools
 * tftp-formula
-* vsftpd-formula
+* vsftpd-formula (optional)
```
aaannz: Please do not specify this as optional. When `retail_branch_init` script is used to configure branch server it expects all formulas to be present.

Bischoff:
@aaannz But I asked on irc channel and someone told me vsftpd was optional, it was only a possible replacement component.

As a matter of fact, the test suite works without it.

What is its real status?

aaannz:
@Bischoff  It is optional to configure, but mandatory to install. It is also part of the suma-retail pattern.

Bischoff:
OK, thanks.

Is it used to download the image? I was under the impression this was done by tftp.

aaannz:
If it is configured, then yes, it is used to download image from branch server to terminal during terminal deployment.

Bischoff:
Thanks for the clarification.

OK, so it is optional in the meaning that you don't need it unless you explicitely want it.

aaannz:
> OK, so it is optional in the meaning that you don't need it unless you explicitely want it.

Yes, exactly.

PR #267, opened 2018-11-21
BZ #1111094 Note about matching FQDN on server/proxy

Loquacity:


adoc/retail_chap_formulas.adoc
```diff
@@ -27,6 +27,8 @@ Once you have made changes to your formula, ensure you apply the highstate to pr
 
 The Bind formula is used to configure the boot image on a terminal.
 
+When you are configuring the bind formula, check that you are using the same fully qualified domain name (FQDN) on both the external and internal branch networks.
+If the FQDN does not match on both networks, the build host will not be recognized as a proxy server.
```
aaannz: I prefer (emphasis on changes):
>When you are configuring the bind formula, check that **Branch server is assigned** the same fully qualified domain name (FQDN) on both the external and internal branch networks.
If the FQDN does not match on both networks, the **Branch server** will not be recognized as a proxy server.

adoc/retail_chap_install.adoc
```diff
@@ -184,9 +184,13 @@ For instructions to install Salt minion see xref:quickstart3_chap_suma_keys_and_
 
 [WARNING]
 ====
+The build host must be a Salt minion.
 Do not install the build host as a traditionally managed client.
 ====
 
+When you are installing the build host, check that you are using the same fully qualified domain name (FQDN) on both the external and internal branch networks.
+If the FQDN does not match on both networks, the build host will not be recognized as a proxy server.
+
```
aaannz: This is not about build host, but branch server. Can you move it down to the `retail.sect.install.branch` section? And adapt the text (emphasis on changes):

> When you are installing the **Branch server with dedicated internal network**, check that you are using the same fully qualified domain name (FQDN) on both the external and internal branch networks.
If the FQDN does not match on both networks, the **Branch server** will not be recognized as a proxy server.

PR #283, opened 2018-12-03
Schedule a channel state for minions if an assigned channel to them is deleted.

admd:
When deleting a channel, channel was being removed correctly from database but in case of minion, channel was still assigned to the system. With this PR, we schedule salt state to update the minion as well.


Loquacity:
Hi @admd thanks for this content! I've just updated the wording a bit. Can you please check that I haven't changed your meaning?

Loquacity:
@admd Thanks for the quick reply! Better now?

Issue #256, opened 2018-10-31
Automatic cleanup of notification messages


renner:
There is a patch currently being reviewed that will introduce a mechanism to cleanup expired notification messages after a configurable lifetime (default still under review, probably 30 or 60 days). This might need to be mentioned in the docs, I guess:

https://github.com/SUSE/spacewalk/pull/6239

renner:
This was merged to `Manager-3.2` meanwhile to be released with the next maintenance update, should be 3.2.4. Notifications lifetime can be configured in `rhn.conf` like this (default lifetime is 30 days):

```
# Configure the lifetime of notification messages in days
java.notifications_lifetime = 30
```

PR #286, opened 2018-12-04
configure notification lifetime

keichwa:
https://github.com/SUSE/doc-susemanager/issues/256
https://github.com/SUSE/spacewalk/issues/6439

back-port to 3.2 when approved.

### DIFF
--- a/adoc/reference-webui-overview.adoc
+++ b/adoc/reference-webui-overview.adoc
@@ -99,7 +99,8 @@ The menu:Main Menu[Home > Notification Messages] page allows you to manage your 
 
 image::home_notification_messages.png[scaledwidth=80%]
 
-You can configure the notifications lifetime in [path]``/etc/rhn/rhn.conf`` (the default lifetime is 30 days):
+The default lifetime of notification messages is 30 days, after which messages are deleted from the database, regardless of read status.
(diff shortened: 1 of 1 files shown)


adoc/reference-webui-overview.adoc
```diff
@@ -99,6 +99,13 @@ The menu:Main Menu[Home > Notification Messages] page allows you to manage your
 
 image::home_notification_messages.png[scaledwidth=80%]
 
+You can configure the notifications lifetime in [path]``/etc/rhn/rhn.conf`` (the default lifetime is 30 days):
```
Loquacity: Hrm. What does "lifetime" mean in this situation? Is it how long the notification remains visible for until it is manually cleared? Or something else? I'd like this to be more specific.

Loquacity: So the wording should probably be something like this:

`You can configure the amount of time before notifications are deleted in [path]``/etc/rhn/rhn.conf`` (the default lifetime is 30 days, after which notifications are deleted from the database, regardless of read status):`

Loquacity: I don't really like the parenthesis, but I'll leave that up to @keichwa to determine, based on context.

keichwa:
thanks for all your input.  please check again.

PR #290, opened 2018-12-06
Added admonition about needing to be registered.

Loquacity:
https://github.com/SUSE/spacewalk/issues/6443


keichwa:
A clickable link back to the beginning of the installation procedure would be nice.  But thinking about all the trouble we have with cross-references, your current solution might be better.

PR #298, opened 2018-12-10
Update reference-webui-admin.adoc

keichwa:
up2date and contents staging
enable commented section again.

### DIFF
--- a/adoc/reference-webui-admin.adoc
+++ b/adoc/reference-webui-admin.adoc
@@ -204,9 +204,8 @@ When settings are done, confirm with clicking the btn:[Update Organization] butt
 
 
 ////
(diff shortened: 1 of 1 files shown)


keichwa:
@renner who can help us with this issue?

renner:
@keichwa: I fear that this section should not be removed yet, at least in the case of traditional clients we are surely still using the `up2date` config file, maybe @mcalmer can confirm this? But I do agree that we should probably at least rename the file at some point.

keichwa:
Johannes Renner <notifications@github.com> writes:

> But I do agree that we should probably at least rename the file at
> some point.

Yes, I also thought renaming it.  Here now is a reminder:
https://github.com/SUSE/spacewalk/issues/6584

-- 
                           90409 Nürnberg, Germany

PR #302, opened 2018-12-17
GPG keys and bootstrapping

Loquacity:
https://github.com/SUSE/spacewalk/issues/6626

Added admonitions about GPG keys in bootstrapping, and general edits.

### DIFF
--- a/adoc/quickstart3_chap_suma_keys_and_first_client.adoc
+++ b/adoc/quickstart3_chap_suma_keys_and_first_client.adoc
@@ -334,7 +334,7 @@ Note that the lock might not deactivate immediately.
 == Registering Salt Clients
 
 There are three possible methods for registering Salt minions.
-You can use a bootstrap repository, a bootstrap script, or install using the the {webui}.
+You can use a bootstrap repository, a bootstrap script, or install using the {webui}.
 
 This section describes using a bootstrap repository.
 Registering Salt clients with a bootstrap client is the same as registering traditional clients, which is described at xref:quickstart3_chap_suma_keys_and_first_client.adoc#registering.clients.traditional[Registering Traditional Clients]. When using this method, ensure you enable the [guimenu]``Bootstrap using Salt`` and activation key options in [guimenu]``Configuration File Deployment``, so that highstate is applied automatically.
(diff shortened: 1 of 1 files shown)


keichwa:
your changes are nice improvements, though!

Loquacity:
> Approved, but it would be good if we'd avoid literally uyuni and sle references. Sometimes {productname} might do the trick. At other places, the ifdef ugliness is needed.

I agree, but from my reading of the issue we're talking specifically about Uyuni tools here? We might need further information from @juliogonzalez on this.

Loquacity:
> your changes are nice improvements, though!

Thanks! This section is ugly though. I have *plans* for the new IA structure on this ;)

adoc/quickstart3_chap_suma_keys_and_first_client.adoc
```diff
@@ -157,23 +157,28 @@ mgr-create-bootstrap-repo -c SLE-15-x86_64 --with-parent-channel sle-product-sle
...
 
-Additionally, it is important that you check the repository information carefully, to ensure it matches the base channel repository.
+It is important that you check the repository information carefully, to ensure it matches the base channel repository.
 If the repository information does not match exactly, the bootstrap script will not be able to download the correct packages.
 
+[IMPORTANT]
+.GPG Keys and Uyuni Client Tools
+====
+The GPG key used by Uyuni Client Tools is not trusted by default.
+When you create your bootstrap script, add the public key fingerprint with the [systemitem]``ORG_GPG_KEY`` parameter.
```
juliogonzalez: What needs to be added to ```ORG_GPG_KEY``` is the filename that contains the public key.

adoc/quickstart3_chap_suma_keys_and_first_client.adoc
```diff
@@ -318,31 +321,38 @@ The feature is not currently supported on {rhel} or Salt clients.
...
 
+You can also use these methods to change existing traditional clients into Salt minions.
 
-The rest of this section assumes you have created a SUSE Manager tools repository.
-You can review creating a tools repository in <<create.tools.repository>>.
+[IMPORTANT]
+.GPG Keys and Uyuni Client Tools
+====
+The GPG key used by Uyuni Client Tools is not trusted by default.
+Either update your bootstrap repository to trust the key explicitly, or use the {webui} to manually trust the key from each client.
```
juliogonzalez: Should we add a suggestion about how to do it? With configuration channels, I guess.

juliogonzalez:
@keichwa I am not sure at all if we should mention Uyuni at the SUSE Manager doc at all or not. That depends, as I guess Uyuni could have, at some point, client tools for distributions or OS that are not supported by SUSE Manager, but still could be interesting.

FMPOV it's a similar situation to what we have for CentOS. It's mentioned at the doc (Best practices), but AFAIK is not officially supported by SUSE.

adoc/quickstart3_chap_suma_keys_and_first_client.adoc
Loquacity: I was OK with leaving it vague just because there are several ways to do it, and I'm fairly certain they're all documented elsewhere.

PR #307, opened 2018-12-20
Updated external links to saltstack docs

Loquacity:
https://bugzilla.suse.com/show_bug.cgi?id=1117497
https://github.com/SUSE/spacewalk/issues/6587

PR #311, opened 2018-12-30
Update per feedback

Loquacity:
Ref https://github.com/SUSE/doc-susemanager/pull/307


Loquacity:
> Fine, thanks!  (With or without colon (but with final dot) is still the
> question.  If with colon I vote to put the https address on a line of
> its own.)

Makes sense. I took the colon out to keep it all as a single sentence.

PR #313, opened 2018-12-31
Add admonition about saltboot parameter conflict.

Loquacity:
https://github.com/SUSE/spacewalk/issues/6640
https://bugzilla.suse.com/show_bug.cgi?id=1119807

### DIFF
--- a/adoc/retail_salt_formulas.adoc
+++ b/adoc/retail_salt_formulas.adoc
@@ -70,8 +70,11 @@ Then you configure one or more partitions:
 image::formula-saltboot-02.png[]
 
 
-////
-////
+[IMPORTANT]
+====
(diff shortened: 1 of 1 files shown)


Bischoff:
" your installation will fail" - maybe not anymore with latest changes from Retail team.

@nadvornik @aaannz does it still fail, or your latest corrections work around that problem?

aaannz:
Once updated packages are submitted, it will no longer fail. It will deploy image and ignore filesystem settings.
The note should be in a meaning that setting image and filesystem values at the same time does not make sense and will not do what customer may expect. (deployed filesystem depends on the selected image and not what customer will set in the saltboot formula)

Bischoff:
OK, so it has to be rephrased accordingly @Loquacity 
* Both parameters should not be entered at same time, as it does not make sense
* If you do enter both parameters, the filesystem settings will be ignored and the image settings will take precedence.

Thanks @aaannz for the update.

PR #330, opened 2019-02-01
Ia rework

jcayouette:
Commented out HA guide content from installation guide. 

Should we just delete this since we publish it from the HA-guide branch?


jcayouette:
Ugh, I need to reclone doc-susemanager seems my origin repo is from a local test again.

PR #332, opened 2019-02-05
2.1 comparison: clarify profile syncing support

moio:
In [this bug report](https://bugzilla.suse.com/show_bug.cgi?id=1124013) we are asked about the reason why `spacecmd`'s `system_syncpackages` does not work with minions.

The reason is that the core of this functionality is not yet ported to Salt.

### DIFF
--- a/adoc/mgr_product_comparison.adoc
+++ b/adoc/mgr_product_comparison.adoc
@@ -51,7 +51,7 @@ This includes features in development and features available only to their paren
 | Package Verification                                                                 | Supported              | Under Review
 | System Locking                                                                       | Supported              | Under Review
 | Configuration File Management                                                        | Supported              | Supported
-| Snapshots and Profiles                                                               | Supported              | Under Review (Profiles are supported)
+| Snapshots and Profiles                                                               | Supported              | Under Review (Profiles are supported, syncing is not)
 | Power Management                                                                     | Supported              | Coming Soon
 |===
 


adoc/mgr_product_comparison.adoc
```diff
@@ -51,7 +51,7 @@ This includes features in development and features available only to their paren
 | Package Verification                                                                 | Supported              | Under Review
 | System Locking                                                                       | Supported              | Under Review
 | Configuration File Management                                                        | Supported              | Supported
-| Snapshots and Profiles                                                               | Supported              | Under Review (Profiles are supported)
+| Snapshots and Profiles                                                               | Supported              | Under Review (Profiles are supported, syncing is not)
```
Loquacity: I would prefer it if we could use "synchronizing" but understand if there's no space in the table.

moio: We use both words in the product:

sync
synchronization

(not to say that this is the right thing to do necessarily, this might as well be a bug! Not sure how the feature is described elsewhere in the documentation)

Loquacity: Yeah, it's one of those ones that the old tech writer in me thinks is "correct" and I know I'm just going to have to let go of eventually. Like I said, if you need to use "sync" because of space restrictions, then go right ahead.

moio:
Please cherry-pick in 3.1 and 3.2 as well!

Loquacity:
Cherry picked to 3.2. Going back to 3.1 is harder, and I'd prefer not to unless it's critical.

PR #338, opened 2019-02-13
improve wording

keichwa:
avoid sentences too long
change wording for clarity (hopefully)


Loquacity:
I liked the original better, although breaking into two sentences is a good idea. "However" is simpler than "nevertheless", and your use of the word "whether" is incorrect in this sentence.

PR #346, opened 2019-02-26
minion-action-cleanup-default

keichwa:
https://github.com/SUSE/spacewalk/issues/6571

### DIFF
--- a/adoc/reference-webui-admin.adoc
+++ b/adoc/reference-webui-admin.adoc
@@ -634,11 +634,10 @@ menu:mgr-sync-refresh-default:[]::
 the default time at which the start of synchronization with SUSE Customer Center (SCC) takes place (``mgr-sync-refresh``).
 
 menu:minion-action-cleanup-default:[]::
-does the following jobs:
-. Completing the unfinished action.
-An unfinished action can occur if the server missed the results of the action.
-. Cleanup of remaining artifacts after action completion.
-For example, it removes script files of corresponding script actions.
+deletes stale minion action data from the file system.
+First it completes the unfinished action.
+An unfinished action can occur if the server has missed the results of the action.
+Then after action completion it removes artifacts such as executed script files.
 
 
 menu:package-cleanup-default:[]::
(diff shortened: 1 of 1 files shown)


Loquacity:
Is there a reason why it's formatted like this, with the single bullets? It seems as though a couple of prose sentences would be clearer.

keichwa:
> Is there a reason why it's formatted like this, with the single bullets? It seems as though a couple of prose sentences would be clearer.

Probably because of implementation details.  It sounds to me as if minion-action-cleanup-default performs two separate actions.  Only if the first one succeeds, the final cleanup will be done.  But I think it's better to change the text.

adoc/reference-webui-admin.adoc
```diff
@@ -633,6 +633,12 @@ Calls the [command]``mgr-register`` command, which synchronizes client registrat
 menu:mgr-sync-refresh-default:[]::
 the default time at which the start of synchronization with SUSE Customer Center (SCC) takes place (``mgr-sync-refresh``).
 
+menu:minion-action-cleanup-default:[]::
+deletes stale minion action data from the file system.
+First it completes the unfinished action.
+An unfinished action can occur if the server has missed the results of the action.
+Then after action completion it removes artifacts such as executed script files.
```
renner: Usually the unfinished actions found here and those that we are deleting artifacts for would not be the same ones. Actions are attempted to be completed only in exceptional cases (we missed the result for some reason), while any successful script execution needs to be cleaned up afterwards (script removal). Maybe we could make it more clear like for example:

"First it tries to complete any possibly unfinished actions by looking up corresponding results (from the Salt job cache). An unfinished action can occur if the server has missed the results of the action. For successfully completed actions it removes artifacts such as executed script files."

keichwa:
https://github.com/SUSE/spacewalk/pull/7163 is somehow related.

Loquacity:
I'm happy with the wording, but we're getting to a point that this should probably be against /beta1 not /develop ...

keichwa:
> I'm happy with the wording, but we're getting to a point that this should probably be against /beta1 not /develop ...

yes, there is already a reminder to myself in the commit msg ;)

keichwa:
I just created https://github.com/SUSE/doc-susemanager/pull/358 -> beta1

PR #344, opened 2019-02-20
DO NOT MERGE: VM display help

cbosdo:
Documentation for PR uyuni-project/uyuni#417
Misses a screenshot to be added

PR #356, opened 2019-03-05
Replicating PR#344 - AT portion

Loquacity:
https://github.com/SUSE/doc-susemanager/pull/344

This is for the portion of the PR that was intended for the AT Guide, which is now in the Admin Guide.


modules/administration/pages/virtualization.adoc
```diff
@@ -463,6 +463,8 @@ Refer to <<sec.virtualization.vmguest-manage.status>> for details.
 
 You can also perform more advanced {vmguest} editing tasks on Salt minions, such as adding or removing disks and network interfaces, and changing the display type.
 
+
+
```
cbosdo: Are these blank lines needed in this commit?

Loquacity: Technically, no, but for readability we like to have three lines of whitespace between sections.

PR #357, opened 2019-03-05
Admin Guide Reorg

Loquacity:


jcayouette:
I do not think spacecmd belongs in the reference book. The ref is for the webui. Spacecmd is a command line utility for advanced users who might wish to script certain actions. I would say this is an administration topic? Your thoughts?

Loquacity:
> I do not think spacecmd belongs in the reference book. The ref is for the webui. Spacecmd is a command line utility for advanced users who might wish to script certain actions. I would say this is an administration topic? Your thoughts?

Just to close the loop  on this one, we discussed this in chat and determined that the spacecmd command reference material should go into the Reference Guide, as a separate section the webui reference material.

Loquacity:
Updated, @jcayouette please re-review :)

PR #371, opened 2019-03-20
Adding mirroring source packages section

Loquacity:
https://github.com/SUSE/spacewalk/issues/5907

### DIFF
--- a/modules/administration/pages/security.adoc
+++ b/modules/administration/pages/security.adoc
@@ -188,6 +188,6 @@ spacewalk-service restart
 Currently, this feature can only be enabled globally for all repositories.
 It is not possible to select individual repositories for mirroring.
 
-Once this feature has been activated, the source packages will become available in the {productname} {webui}.
+When this feature has been activated, the source packages will become available in the {productname} {webui}.
 They will be shown as sources for the binary package, and can be downloaded directly from the {webui}.
 Source packages cannot be installed on clients using the {webui}.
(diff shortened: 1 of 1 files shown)


modules/administration/pages/security.adoc
```diff
@@ -164,3 +164,30 @@ $> rpm --import http://suma-refhead-srv.mgr.suse.de/pub/company.key
 ====
 For salt managed systems it might make sense to use a state to trust GPG keys.
 ====
+
+
+== Mirror Source Packages
+
+If you build your own packages locally, or if you require the source code for your packages for legal reasons, it is possible to mirror the source packages on your local system.
```
mcalmer: 'Local system' sounds too generic. I would say SUSE Manager (Server)

modules/administration/pages/security.adoc
```diff
@@ -164,3 +164,30 @@ $> rpm --import http://suma-refhead-srv.mgr.suse.de/pub/company.key
...
+. Restart the spacewalk service to pick up the changes:
+----
+spacewalk-service restart
+----
+
+Currently, this feature can only be enabled globally for all repositories.
+It is not possible to select individual repositories for mirroring.
+
+Once this feature has been activated, the source packages will become available in the {productname} {webui}.
+They will be shown as sources for the binary package, and can be downloaded directly from the {webui}.
```
mcalmer: I need to check, but maybe we have also an API endpoint to get the package.

Loquacity: We can do another patch once you have that info.

PR #392, opened 2019-03-28
Remove step containing links to download.opensuse.org

Loquacity:


modules/client-configuration/pages/clients-ubuntu.adoc
```diff
@@ -52,7 +52,7 @@ Some actions are not yet supported:
 
 This section describes how to manually prepare {ubuntu} clients for registration to your {susemgr} Server and synchronize the repositories.
 
-
+////
 .Procedure: Preparing an {Ubuntu} 16.04 Client for Registration
```
juliogonzalez: Shouldn't you remove the procedure for 16.04 if you are going to have a single one for both versions?

juliogonzalez: Ah... this is a comment, right?

Loquacity: I just commented it out until we're happy. I'm allergic to deleting content until I'm 110% sure ;)

PR #406, opened 2019-04-04
SLES11 Salt client SSL error

Loquacity:
Added troubleshooting section for this issue.

https://bugzilla.suse.com/show_bug.cgi?id=1131381
https://github.com/SUSE/spacewalk/issues/7491


modules/client-configuration/pages/manual-registration/salt-registration.adoc
```diff
@@ -110,3 +110,28 @@ You can adjust this manually to correct the error and register the cloned system
 
 
 For more information and instructions, see pass:c[xref:FILENAME.adoc#bp.chapt.suma3.troubleshooting.registering.cloned.salt.systems[]].
+
+
+=== SSL errors
+
+On {slsa}{nbsp}11 systems, salt clients can sometimes have SSL errors which make some operations unusable, including package management and bootstrapping.
```
mcalmer: This is not only happening with salt minions, traditional clients are affected as well and even more as the registration already fail with this error.

As salt itself does not use any SSL i think it would be better to just say any SLES11 client.

modules/client-configuration/pages/manual-registration/salt-registration.adoc
```diff
@@ -110,3 +110,28 @@ You can adjust this manually to correct the error and register the cloned system
...
+----
+Repository 'SLES11-SP4-SUSE-Manager-Tools x86_64' is invalid.
+[|] Valid metadata not found at specified URL(s)
+Please check if the URIs defined for this repository are pointing to a valid repository.
+Skipping repository 'SLES11-SP4-SUSE-Manager-Tools x86_64' because of the above error.
+Download (curl) error for 'www.example.com':
+Error code: Unrecognized error
+Error message: error:1409442E:SSL routines:SSL3_READ_BYTES:tlsv1 alert protocol version
+----
+
```
mcalmer: I think it would be good to explain that apache in SUMA 4.0 is only allowing TLSv1.2 by default.
But SLES11 is so old that it does not support this protocol version.
To make SLES11 work with SUMA 4.0 the customer have to relax the protocol requirements.

modules/client-configuration/pages/manual-registration/salt-registration.adoc
Loquacity: In that case, I have this in the wrong section.

Loquacity: Moved all the troubleshooting bits into their own section.

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -0,0 +1,63 @@
...
+Repository 'SLES11-SP4-SUSE-Manager-Tools x86_64' is invalid.
+[|] Valid metadata not found at specified URL(s)
+Please check if the URIs defined for this repository are pointing to a valid repository.
+Skipping repository 'SLES11-SP4-SUSE-Manager-Tools x86_64' because of the above error.
+Download (curl) error for 'www.example.com':
+Error code: Unrecognized error
+Error message: error:1409442E:SSL routines:SSL3_READ_BYTES:tlsv1 alert protocol version
+----
+
+This occurs because Apache requires TLS{nbsp}v1.2, but older versions of {slsa} do not support this version of the TLS protocol.
```
mcalmer: RHEL 6 might have the same problem. Maybe QA can test it when there is a bit of time after Beta 2 is out.

Loquacity: We can always update if they find new info.  I'll merge this now.

PR #408, opened 2019-04-05
BZ1131398 - Client Registration 

Loquacity:
https://bugzilla.suse.com/show_bug.cgi?id=1131398
https://github.com/SUSE/spacewalk/issues/7493


modules/client-configuration/pages/manual-registration-overview.adoc
```diff
@@ -0,0 +1,11 @@
+[[manual-overview]]
+= Client Registration Overview
+
+There are two ways to register clients to your {productname} Server.
+Both methods work for either traditional or salt clients.
```
mcalmer: The web ui variant work only for salt clients!

Traditional must be registered via bootstrap script.

modules/client-configuration/pages/registration-bootstrap.adoc
```diff
@@ -70,32 +62,32 @@ The bootstrap script is also available at [path]``https://example.com/pub/bootst
 In this section you will copy and modify the template bootstrap script you created from pass:c[xref:FILENAME.adoc#generate.bootstrap.script[]].
 
 A minimal requirement when modifying a bootstrap script for use with {productname} is the inclusion of an activation key.
-Depending on your organizations security requirements it is strongly recommended to include one or more (GPG) keys (for example, your organization key, and package signing keys).
-For this tutorial you will be registering with the activation keys created in the previous section.
+Depending on your organization's security requirements it is strongly recommended to include one or more GPG keys (for example, your organization key, and package signing keys).
```
mcalmer: This phrase is not really correct. It is not about "organization's security requirements".

If you try to install a signed package where the system does not "trust" the signing key, it will ask the user if he really want to install that package.

But as we talk here about "non-interactive" processes, the default answer to this question is "no".
So not having the GPG keys trusted on the system prevent installing and updating packages.

By default every system trust the vendor GPG key. As long as you install only packages from vendor you do not need to trust additional keys. But when you start to install packages signed with your own key or RPMs from a different vendor (e.g. SUSE Client Tools packages on RedHat or Ubuntu) you need to trust the GPG keys which were used to sign these packages.

modules/client-configuration/pages/registration-bootstrap.adoc
```diff
@@ -29,15 +21,15 @@ When you create your bootstrap script, add a path to the file containing the pub
...
 ====
 
 This procedure describes how to generate a bootstrap script.
 
 [[create.boot.script]]
 .Procedure: Creating a Bootstrap Script
-. From the {productname} {webui}, browse to menu:Main Menu[Admin > Manager Configuration > Bootstrap Script]. For more information, see pass:c[xref:FILENAME.adoc#s3-sattools-config-bootstrap[]].
-. In the [guimenu]``SUSE Manager Configuration - Bootstrap`` dialog disable [guimenu]``Bootstrap using Salt``.
+. In the {productname} {webui}, navigate to menu:Admin[Manager Configuration > Bootstrap Script].
+. In the [guimenu]``SUSE Manager Configuration - Bootstrap`` dialog, disable [guimenu]``Bootstrap using Salt``.
```
lkotek: I think we should mention both ways, disabling `Bootstrap using Salt` is necessary only when registering as traditional client.

modules/client-configuration/pages/manual-registration-overview.adoc
Loquacity: Oof, got it.

modules/client-configuration/pages/registration-bootstrap.adoc
Loquacity: Trying to capture that in a short sentence is hard. How's this?

> Most packages are signed with GPG, so you will also need to have trusted GPG keys on your system to install them.

mcalmer: Sounds good.

mcalmer:
Seems no changes has been pushed.

Loquacity:
> Seems no changes has been pushed.

Turns out you need to push the commit after you've created it! Who knew?!

PR #416, opened 2019-04-08
No registration code required for SLE15SP1 and SUSE Manager 4.0

juliogonzalez:
Required for Beta3.

Does not apply to Uyuni.

### DIFF
--- a/modules/installation/pages/general-requirements.adoc
+++ b/modules/installation/pages/general-requirements.adoc
@@ -47,6 +47,10 @@ Depending on your organization's setup, you might also need to activate your sub
 
 From {slsa}{nbsp}15{nbsp}SP1, {susemgr} Server and Proxy is available as a base product, and can be installed with the {slsa} Unified Installer.
 
+ifeval::[{suma-content} == true]
+Therefore you do not require a separate registration code for {slsa} {sles-version} {sp-version}, only for Server or Proxy
+endif::[]
+
 Download {slsa}{nbsp}15{nbsp}SP1 or higher from https://www.suse.com/products/server/download/[SUSE Linux Enterprise Server - Media Download]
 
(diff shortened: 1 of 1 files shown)


modules/installation/pages/general-requirements.adoc
```diff
@@ -47,6 +47,10 @@ Depending on your organization's setup, you might also need to activate your sub
 
 From {slsa}{nbsp}15{nbsp}SP1, {susemgr} Server and Proxy is available as a base product, and can be installed with the {slsa} Unified Installer.
```
juliogonzalez: Unrelated, but should that use `{slsa} {nbsp}{sles-version}{nbsp} {sp-version}`?

modules/installation/pages/general-requirements.adoc
```diff
@@ -47,6 +47,10 @@ Depending on your organization's setup, you might also need to activate your sub
...
 
+ifeval::[{suma-content} == true]
+Therefore you do not require a separate registration code for {slsa} {sles-version} {sp-version}, only for Server or Proxy
+endif::[]
+
 Download {slsa}{nbsp}15{nbsp}SP1 or higher from https://www.suse.com/products/server/download/[SUSE Linux Enterprise Server - Media Download]
```
juliogonzalez: Same as before (for a separate PR)

modules/installation/pages/general-requirements.adoc
Loquacity: I wrote it that way because we don't want that version to change. Even after we're based on a newer version, it will still be true that it's a base product from 15 SP1.

juliogonzalez: That's right, but my understanding is that those values will always come from the yaml files.

@jcayouette ?

PR #420, opened 2019-04-11
Fix build errors

Loquacity:
Fixed those pesky build errors in the Install Guide, and updated *-site.yml to point to master

### DIFF
--- a/modules/installation/pages/server-setup.adoc
+++ b/modules/installation/pages/server-setup.adoc
@@ -2,13 +2,6 @@
 = SUSE Manager Setup
 
 
-
-
-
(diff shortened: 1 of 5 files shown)


modules/installation/pages/server-setup.adoc
```diff
@@ -146,7 +136,7 @@ This can take a long time as {productname} will copy channel software sources fr
 
 
 [TIP]
-.PostgreSQL and Transparant Huge Pages
+.PostgreSQL and Transparent Huge Pages
```
keichwa: good catch!

PR #424, opened 2019-04-11
Live Patching

Loquacity:
* Split Live Patching chapter into three: intro, SLES 15, SLES12
* Edited the entire existing chapter to improve procedures.
* For SLES 15 chapter, updated from kgraft to livepatch
* Moved an admonition about channels to the channel management section.


modules/administration/pages/live-patching-sles12.adoc
```diff
@@ -0,0 +1,113 @@
...
+
+.Procedure: Setting up for Live Patching
+
+. Select the client you want to manage with Live Patching from menu:Systems[Overview], and navigate to the menu:Software[Packages > Install] tab.
+Search for the [systemitem]``kgraft`` package, and install it.
++
+image::enable_live_patching_kgraft_install.png[scaledwidth=80%]
+. Apply the highstate to enable Live Patching.
+. Repeat for each client that you want to manage with Live Patching.
+. To check that Live Patching has been enabled correctly, select the client from menu:Systems[Systems List], and ensure that [systemitem]``Live Patching`` appears in the [guimenu]``Kernel`` field.
```
cbbayburt: IIRC, systems will need a reboot when live patching is enabled for the first time. I need to test it out in both versions to verify. If that is correct, we should mention it here.

I'll get back to you after I try it out ASAP.

Loquacity:
> As a first step I think this is ok.
> 
> There is 1 point I wanted to note, maybe we should create new issues for it.
> 
> We describe here `spacewalk-manage-channel-lifecycle` tool which was the only option in the past.
> But now @ncounter with naica implemented "Content Lifecycle Management" feature which could do the same.
> Before we describe it, we need to be sure that the required filtering is working.

Yes, I was wondering that as I was editing (given that I only just wrote the content lifecycle stuff, so it's fresh in my mind). Probably best to do that in a new issue at this stage, though, just to keep things organised. Also, because we haven't got content lifecycle management fully up and running yet.

mcalmer:
@Loquacity created : https://github.com/SUSE/spacewalk/issues/7588

modules/administration/pages/live-patching-sles12.adoc
Loquacity: @cbbayburt Any update here?

cbbayburt: @Loquacity, I can confirm for both versions that a reboot is required after the **initial** install of a live patching package.
In the UI Overview page, you can see both the live patching version and the reboot required text.

PR #425, opened 2019-04-11
updated makefile for uyuni/suma targets

jcayouette:
Adjusted build variables for both SUMA and UYUNI outputs. This means we no longer have to fiddle with these before selecting a target.

TODO: We need to adjust antora.yml at build time depending on the productname_uyuni or productname_suma.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,30 +3,48 @@
 # Inspired/modified from Owncloud's documentation Makefile written by Matthew Setter
 SHELL = bash
 
-# Productname to replace
-PRODUCTNAME ?= 'SUSE Manager'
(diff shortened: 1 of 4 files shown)


jcayouette:
```
# Uncomment name and title to build for SUSE Manager
name: suse-manager
title: SUSE Manager

# Uncomment name and title to build for Uyuni
#name: uyuni
#title: Uyuni

# Which branch is the default for this playbook.
version: master

# Uncomment one of the following home pages depending on your target (SUMA or Uyuni)
start_page: ROOT:index-suma
# start_page: ROOT:index-uyuni
```

PR #438, opened 2019-04-24
Minion id is visible in System Info box

ncounter:
See https://github.com/uyuni-project/uyuni/pull/883

### DIFF
--- a/modules/reference/pages/systems/system-details.adoc
+++ b/modules/reference/pages/systems/system-details.adoc
@@ -71,7 +71,7 @@ IPv6 Address:::
 The IPv6 address of the client.
 
 Minion Id:::
-If the client is a salt minion, the minion id value is shown.
+On salt clients only, shows the minion identification value.
 
 Virtualization:::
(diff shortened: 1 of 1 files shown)


modules/reference/pages/systems/system-details.adoc
```diff
@@ -70,6 +70,9 @@ The IP address of the client.
 IPv6 Address:::
 The IPv6 address of the client.
 
+Minion Id:::
+If the client is a salt minion, the minion id value is shown.
```
Loquacity:
```suggestion
For salt clients, the minion identification value is shown.
```

Loquacity:
```suggestion
On salt clients only, shows the minion identification value.
```

ncounter: That's the difference outcome when *you* are a mother language and a great writer......but I don't!! :wink:

Loquacity: I still love your work, Dario <3

PR #444, opened 2019-04-29
Add notes about bandwidth usage at registration time

moio:
Adding some notes on bandwitdh usage during registration depending on the Activation Key.

General context: https://github.com/SUSE/spacewalk/wiki/Proxy-NG-research


moio:
@joesusecom, @abhinav-puri: this is more or less all I had in mind for point 3 in the Wiki page above. Please let me know in case there is something to add.

moio:
@jcayouette I hope this is the right place to add such notes!

Loquacity:
I  think this location is fine. Writing is acceptable, but I would like to revisit this whole section once this has merged, it could use a polish.

keichwa:
> I think this location is fine. Writing is acceptable, but I would like to revisit this whole section once this has merged, it could use a polish (opened [SUSE/spacewalk#7711](https://github.com/SUSE/spacewalk/issues/7711) so I don't forget).

:+1: Thanks for your time, @moio !

PR #445, opened 2019-04-30
Salt Guide Updates

Loquacity:
- [x] General Structure updates
- [x] Content from https://github.com/SUSE/spacewalk/issues/7652

This will also need a broad edit across the whole doc, but it's getting long enough as it is. Will do in another PR once this merges.


modules/salt/pages/salt-intro.adoc
```diff
@@ -2,19 +2,11 @@
 = Introduction
 
 
+Salt is a configuration management system used by {productname} to manage clients.
+Since {productname}{nbsp}3, you have the option of registering salt clients.
```
keichwa:
```suggestion
You have the option of registering Salt clients.
```

Loquacity: So I was trying to include the information about it being available in SUSE Manager 3+, but got  stuck on wording. Suggestions? Or maybe it's not important which version it was introduced  in?

keichwa:

> So I was trying to include the information about it being available in
> SUSE Manager 3+, but got stuck on wording. Suggestions? Or maybe it's
> not important which version it was introduced in?

I think in the context of uyuni/suma4, the version number is no longer
important.

Maybe, we should stress that you can register a Salt client at(?) the
{productname} server and that the Salt master runs on the {productname}
server.

IIRC, the Salt master _must_ even run on the {productname} server.  Not
sure whether that's feature or a limitation...

[Thanks for working on these details!  Of course, many things of this
architecture are not that clear to me!]

-- 
                           90409 Nürnberg, Germany

Loquacity:
> I think in the context of uyuni/suma4, the version number is no longer important. Maybe, we should stress that you can register a Salt client at(?) the {productname} server and that the Salt master runs on the {productname} server.

Yep, I like that idea. Will update.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
...
+You can use these rate-limiting features to control the load in your environment.
+
+These features are all configured in the [filename]``/etc/rhn/rhn.conf`` configuration file.
+
+
+
+== Batch Size
+
+Actions are processed in batches.
+This parameter sets the maximum number of clients that can execute a single action in each batch.
```
moio: Not really.

Batching and rate limiting are different concepts, and what Salt/`{productname}` does is rate limiting, not batching. It's unfortunate that this very feature is named Batching in Salt (upstream) and existed long before we added support in `{productname}`, thus we had to adopt its jargon.

Example: a command has to be executed on 100 minions, and the batch size is 20.

**Batching** means:
 1. that 5 sets of 20 minions each are created
 1. that initially, 20 commands are issued (one per minion in batch 1)
 1. that as soon as all 20 minions in batch 1 are done, batch 2 begins and so on
 1. thus, that at any point in time there can only be 20 or less commands in execution
 1. typically the number will be 20 when batches are started and will gradually diminish until batch end. On average, less than 20 will be active

**Rate limiting** means:
 1. that only 1 set of 20 minions is created (**unlike** in batching)
 1. that initially, 20 commands are issued (one per minion in the only one batch) (**same** as in batching)
 1. that as soon as any minion is done, another minion immediately* starts (**unlike** in batching)
    - corner case: more than one minion is done precisely at the same time. An equal number of minions start
 1. thus, that at any point in time there can only be 20 or less commands in execution (**same** as in batching)
 1. typically the number will be exactly 20, give or take a few during issuing. On average, the number of active commands will be very close to 20 (**unlike** in batching)

*Note the asterisk at point 3. More about that later*

Summarizing: both batching and rate limiting guarantee no more than 20 commands are active at any given point in time, but rate limiting does a much better job in keeping the number close to 20 for the whole duration, resulting in better performance. That's why the latter is implemented and not the former!


Hopefully the SUSECON presentation I sent you (in particular the animations) make this clearer.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Choosing a lower number can prevent system thrashing, but might reduce overall performance as parallelization is reduced.
```
moio: While the word thrashing probably conveys the meaning intuitively, it has a very specific meaning in jargon and it does not really apply to this case:

https://en.wikipedia.org/wiki/Thrashing_(computer_science)

I'd be happier with something like "Choosing a lower number will reduce the parallelism and load, but potentially also reduces Action processing performance."

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Adjust the [systemitem]``java.salt_batch_size`` parameter.
+Defaults to 200.
```
moio: @chiaradiamarcelo are we set on 200? Maybe we want to follow the traditional client default and change that to 100? I can't recall the final decision on this.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+
+
+== Batch Delay
+
+This parameter sets the amount of time, in seconds, to wait after a batch is processed before beginning to process the next batch.
```
moio: > 3. that as soon as any minion is done, another minion immediately* starts (**unlike** in batching)
>    - corner case: more than one minion is done precisely at the same time. An equal number of minions start

*: this parameter sets a certain delay after one or more minions are done and the time new command(s) are issued to an equal number of minions.

Why would one want to add such a delay, instead of starting immediately? Doesn't that harm performance?

It might not, as starting immediately would probably mean (in the example above) that the command is initially sent to 20 minions all at once, then 1 is done and 1 is issued, then 1 is done and 1 is issued, then 1 is done and 1 is issued... 80 times. For a grand total of 81 issued commands.

If we set a batch delay of, say, 5 seconds, chances are that several minions might complete before the next command is issued, thus giving Salt the opportunity to collect the results and issue less commands, For example I can imagine a first batch of 20, then we get 10 minion results before the 5 seconds elapse so we issue one command to 10 target minions, then we get 12, then maybe 8... The grand total could be 8-10 commands, much less than the case above.

This is good because of how Salt implements broadcasting to minions (many commands each targeting one minion are much less efficient than a few targeting big groups).

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Adjust the [systemitem]``java.salt_batch_delay`` parameter.
+Defaults to 1.0 seconds.
+
+
+
+== Ping Timeout
+
+During Salt batch calls, Salt clients will perform a presence ping.
+This parameter changes the timeout setting for the presence ping, in seconds.
```
moio: Minions not responding within the presence ping timeout will be excluded from the command and will return an error.

modules/salt/pages/salt-rate-limiting.adoc
chiaradiamarcelo: As for now, the default parameter is 200, yes

modules/salt/pages/salt-rate-limiting.adoc
chiaradiamarcelo: Actually they will not return an error, they will just be identified as "not available" minions, they will be excluded form the call as said, and in the UI we show this with a "Minion is down" message.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Adjust the [systemitem]``java.salt_batch_size`` parameter.
+Defaults to 200.
+
+
+
+== Batch Delay
+
+This parameter sets the amount of time, in seconds, to wait after a batch is processed before beginning to process the next batch.
+The first batch will process the number of actions determined in the batch size parameter, and subsequent batches will process fewer actions.
```
chiaradiamarcelo: I find this part a little bit confusing, given that we are talking about actions, but in reality what happens is that we talk about minions. 
The first batch actually will process whatever we say in a "batch size" amount of minions, and subsequent batches will process likely fewer minions than the first one, but at the end we will always have, at any given point in time, the same number of minions in execution in parallel.

modules/salt/pages/salt-rate-limiting.adoc
chiaradiamarcelo: Edit, the default value at the moment is 100, same as the trad clients

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
+[[salt.rate.limiting]]
+= Rate Limiting
+
+Salt is able to run many commands in parallel on a large number of clients.
+This can potentially create large amounts of load on your infrastructure.
+You can use these rate-limiting features to control the load in your environment.
```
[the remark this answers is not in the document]
Loquacity: No, I think you're right. What I mean there is parameters.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+These features are all configured in the [filename]``/etc/rhn/rhn.conf`` configuration file.
+
+
+
+== Batch Size
+
+Actions are processed in batches.
```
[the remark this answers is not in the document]
Loquacity: Also added a note to that effect.

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: Yes, the presentation (and your animations!) were very helpful. I wasn't aware of this specific difference, though.

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: Hrm, I thought thrashing was a more common way of saying this, I didn't realise how specific it was. I like your suggestion.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,60 @@
...
+[NOTE]
+====
+Salt commands that are executed from the command line are not subject to these parameters.
+====
+
+
+
+== Batch Size
+
+This parameter sets the maximum number of clients that can execute a single action at any one time.
```
chiaradiamarcelo: maybe you meant "...at any given point in time." or "...at the same time." ?

Loquacity:
> Changes look good. Thanks for the updates. One thing. You do not need to adjust the entities file contained in the partials directory. Have a look at **suma-site.yml** :)
> 
> We set our entities globally in the antora catalog when we build. This means we only need to maintain the single files for uyuni/suma
> 
> "My bad btw, I have yet to remove the partials folders. We no longer need these."

OK, thanks for clearing that up. I was getting build errors on the Install Guide partials entity file earlier, so I was getting confused about how they work.

Loquacity:
@moio I need your review to merge, can you please take another look for me? Thanks!

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,60 @@
...
+Defaults to 4 seconds.
+
+
+
+== Job Timeout
+
+During Salt batch calls, a job runs to handles the presence ping from the clients.
+This parameter changes the timeout setting for gathering the presence ping, in seconds.
+
+If a minion does not reply to a ping within the allocated time, it will be marked as [systemitem]``not available``, and will be excluded from the command.
```
moio: This note is basically a conclusion of both the `salt_presence_ping_timeout` and the `salt_presence_ping_gather_job_timeout`.

What happens is well described here: https://www.suse.com/documentation/suse-manager-3/book_suma_best_practices_3/data/bp_chap_salt_minion_scaleability_timeouts.html

I think it is worth to put that detailed description here as well - otherwise it's hard to understand why there are two distinct timeout settings and what their mean is.

Note that the logic there about `timeout`/`gather_job_timeout` (total default: 240 seconds) is the same applied for `salt_presence_ping_timeout`/`salt_presence_ping_gather_job_timeout` (total default: 5 seconds) - only that the latter couple only affects presence pings. We want to be able to set those differently to more quickly detect minions that are down while also allowing for much wider margins for non-ping commands.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,60 @@
...
+Increasing the delay increases the chance that multiple minions will have completed before the next action is issued, resulting in fewer overall commands, and reducing load.
+
+Adjust the [systemitem]``java.salt_batch_delay`` parameter.
+Defaults to 1.0 seconds.
+
+
+
+== Ping Timeout
+
+During Salt batch calls, Salt clients will perform a presence ping.
```
chiaradiamarcelo: Maybe we could say "At the beginning" instead of "During", given that that's the way it works.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,60 @@
+[[salt.rate.limiting]]
+= Rate Limiting
+
+Salt is able to run commands in parallel on a large number of minions.
+This can potentially create large amounts of load on your infrastructure.
+You can use these rate-limiting parameters to control the load in your environment.
```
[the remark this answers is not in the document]
Loquacity: Yeah, I went back and  forth on this a bit, and consulted with the Write the Docs community to  understand what 'best practice' is here. Consensus formed around the distinction between using two words when using it as a  noun or a verb, and a hyphen when using it as an adjective.

modules/salt/pages/salt-rate-limiting.adoc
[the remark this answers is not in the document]
Loquacity: This is awkward phrasing (the original isn't much better, tbh). Assuming the master pings the minion, which responds, this wording might be better:

```suggestion
During Salt batch calls, the Salt master pings the minions to ensure they are still present.
```

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,60 @@
...
+[NOTE]
+====
+Salt commands that are executed from the command line are not subject to these parameters.
+====
+
+
+
+== Batch Size
+
+This parameter sets the maximum number of clients that can execute a single action at the same time.
```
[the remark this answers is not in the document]
Loquacity: I don't think there's a good argument for initcaps here. Like I said in another review, though, our docs are littered with initcaps, so I doubt there's a consensus here.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,60 @@
...
+This parameter changes the timeout setting for the presence ping, in seconds.
+
+Adjust the [systemitem]``java.salt_presence_ping_timeout`` parameter.
+Defaults to 4 seconds.
+
+
+
+== Job Timeout
+
+During Salt batch calls, a job runs to handles the presence ping from the clients.
```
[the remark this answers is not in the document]
Loquacity: Same as above.

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: OK, what about  this?

```suggestion
Salt batch calls begin with the Salt master performing a presence ping on the minions.
```

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: This content exists in the client cfg guide now, but I think it's better suited to the Salt Guide. I'll move it in.

modules/salt/pages/salt-rate-limiting.adoc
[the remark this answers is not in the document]
moio: Very interesting, thanks for sharing, will try to adhere to this rule from now on.

modules/salt/pages/salt-rate-limiting.adoc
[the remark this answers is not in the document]
moio: Your suggested wording conveys the message better, I like it.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+== Ping Gather Job Timeout
+
+During Salt batch calls, a job runs to handles the presence ping from the clients.
+This parameter changes the timeout setting for gathering the presence ping, in seconds.
+
+If a minion does not reply to a ping within the allocated time, it will be marked as [systemitem]``not available``, and will be excluded from the command.
+The {webui} will show a [systemitem]``minion is down`` message in this case.
+
+For more information on minion timeouts, see xref:scale-minions.adoc[].
```
moio: I suggest to:
 - merge the "Ping Timeout" and "Ping Gather Job Timeout" sections under one title
 - start with "During Salt batch calls, the Salt master will perform a presence ping to target minions."
 - explain the presence ping purpose
 - continue with something like:

"As an ordinary Salt command, the presence ping is subject to the same kind of timeouts that affect all Salt commands, namely the main `timeout` and the `gather_job_timeout` - their meaning in detail is explained at `xref...`.

However, the default lengths of these timeouts are significantly shorter for presence pings, to allow quicker detection of unresponsive minions. They can also be configured individually, via the following parameters in `rhn.conf`..."

modules/salt/pages/salt-rate-limiting.adoc
chiaradiamarcelo: Nice, thanks

modules/salt/pages/salt-rate-limiting.adoc
chiaradiamarcelo: We can also include, in the "presence ping purpose" section, that the execution of the batching calls will start when either all the minions replied before the presence ping timeout, or we actually reach the time out and we execute it with the minions that  reply. What happens first

modules/salt/pages/salt-rate-limiting.adoc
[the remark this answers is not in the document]
Loquacity: No problem. There's quite a few words that work like that in IT.

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: OK, I've combined the batching bit as well, and made them more or less parallel. Think I  got everything!

Loquacity:
Re-requesting reviews from  @keichwa and @jcayouette because there's been some significant wording changes since they last looked at this.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+Salt commands that are executed from the command line are not subject to these parameters.
+====
+
+
+
+== Batching
+
+There are two parameters that control how actions are sent to clients, one for the batch size, and one for the delay.
+
+When the Salt master sends a batch of actions to the target minions, it will send it to the number of minions determined in the batch size parameter.
```
moio: Hmm, technically all the Salt master sends are commands or function calls. Actions are a `{productname}`-specific concept. Given that batching is a `{productname}`-specific concept too, I would reword that as:
```suggestion
When the {productname} Server executes an action against multiple target minions, it will send a Salt command to the number of minions determined in the batch size parameter.
```

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+As each minion completes, the action begins on a new minion after the specified delay period.
```
moio:
```suggestion
As each minion completes, the command begins on a new minion after the specified delay period.
```

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+Choosing a lower batch size will reduce system load and parallelism, but might reduce overall performance for processing actions.
+
+The batch size parameter sets the maximum number of clients that can execute a single action at the same time.
```
moio:
```suggestion
The batch size parameter sets the maximum number of minions that can execute a single action at the same time.
```

It's important to specify minions here, as traditional clients will not be affected by this parameter.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+Adjust the [systemitem]``java.salt_batch_size`` parameter.
+Defaults to 100.
+
+Increasing the delay increases the chance that multiple minions will have completed before the next action is issued, resulting in fewer overall commands, and reducing load.
```
moio:
```suggestion
Increasing the delay increases the chance that multiple minions will have completed before the next command is issued, resulting in fewer overall commands, and reducing load.
```

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+The batch delay parameter sets the amount of time, in seconds, to wait after an action is processed before beginning to process the action on the next minion.
```
moio:
```suggestion
The batch delay parameter sets the amount of time, in seconds, to wait after a command is processed before beginning to process the command on the next minion.
```

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+== Presence Ping Timeout
+
+There are two parameters that control how presence pings from the Salt master are handled, one for the ping timeout, and one for the ping gather job.
+
+Salt batch calls begin with the Salt master performing a presence ping on the target minions.
+A ping gather job runs on the Salt master to handle the incoming pings from the minions.
+Batched actions will begin only after all minions have either responded to the ping, or timed out.
+
+The presence ping is an ordinary Salt command, so it is subject to the same timeout parameters as all other Salt commands.
```
moio: Hmmm, not exactly. I don't know how to put this in understandable English, but in my tech-English it sounds like "it is subject to the same *kind* of timeout parameters, albeit with different values".

In concrete terms: `salt_presence_ping_timeout` and `salt_presence_ping_gather_job_timeout` override `timeout` and `gather_job_timeout` respectively, but only for presence pings.

More below.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+You can configure the global timeout values in the [filename]``/etc/salt/master.d/custom.conf`` configuration file.
+However, to allow for quicker detection of unresponsive minions, the timeout values for presence pings can be significantly shorter than those used elsewhere.
+Therefore, we recommend that you override the global timeout values by setting specific values for presence pings in [filename]``/etc/rhn/rhn.conf``.
```
moio: I think this needs to be clarified a bit more.

`salt_presence_ping_timeout` and `salt_presence_ping_gather_job_timeout` are *by default* set much lower than the global counterparts. We do not expect the majority of users to want to change these values.

By default, in `{productname}`:
* `timeout`: 2 minutes
* `gather_job_timeout`: 2 minutes
* total timeout for general Salt commands from `{productname}`: `timeout` + `gather_job_timeout` = 4 minutes
* `salt_presence_ping_timeout`: 1 second
* `salt_presence_ping_gather_job_timeout`: 4 seconds
* total timeout for presence pings: `salt_presence_ping_timeout` + `salt_presence_ping_gather_job_timeout` = 5 seconds

Users will rarely want to set those even lower.

The lower the total presence ping timeout is, the higher the chance for false negatives: minions reported as non-responding while they are actually responding, they might just be too busy (or too slow for whatever reason) to respond quickly enough.

The higher the total presence ping timeout is, the more accurate the test will be: most minions, even if slow, will have a chance to respond to the presence ping before timeouts elapse. OTOH, commands will start later and this could limit throughput when targeting a large set of minions some of which are slow.

Note that if all minions are fast, the choice of presence ping timeouts is irrelevant!

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+
+
+== Presence Ping Timeout
+
+There are two parameters that control how presence pings from the Salt master are handled, one for the ping timeout, and one for the ping gather job.
+
+Salt batch calls begin with the Salt master performing a presence ping on the target minions.
+A ping gather job runs on the Salt master to handle the incoming pings from the minions.
+Batched actions will begin only after all minions have either responded to the ping, or timed out.
```
chiaradiamarcelo: "Batched commands" instead of "Batched actions"

modules/salt/pages/salt-rate-limiting.adoc
chiaradiamarcelo: Not sure if it worth to mention it, but I found this a little bit tricky.
"As each minion completes, the command begins on a new minion after the specified delay period" is true but does not reflect how it behaves completely. This to me gives me a feeling of "the commands on the minions will be executed sequentially as each minion completes".

What actually happens is that, as the minions complete, a next batch is gonna be scheduled to execute the command in new minions. The size of that new batch will be equal to the number of minions that completed the command at the moment the "delay" time expired. Or, better said, the empty places that are left at that moment. (Sorry Lana, I'm not good with rewording as you can see haha)

modules/salt/pages/salt-rate-limiting.adoc
chiaradiamarcelo: see my comment above in https://github.com/SUSE/doc-susemanager/pull/445/files#r282799570

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: Heh, it's fine. I've had another go :)

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: I had interpreted `timeout` and `gather_job_timeout` as global variables, and the  presence_ping  ones as specific variables that override the  global ones. Am I misunderstanding?

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: OK, I've added that bit at the end. I can't really tell what changes you need in lines 47-50.

modules/salt/pages/salt-rate-limiting.adoc
moio: That is correct. Let me give a try to rephrase the sentence above to try to show what I wanted to say:

```suggestion
The presence ping is an ordinary Salt command, but is not subject to the same timeout parameters as all other Salt commands (`timeout`/`gather_job_timeout`), rather, it has its own parameters (`presence_ping_timeout`/`presence_ping_gather_job_timeout`).
```

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: After a call @moio 2019-05-14 I think we have this resolved now.

Loquacity:
> Added another few remarks. I think it is important to fully explain the mechanism at this point, or to put a link to the fuller explanation if its natural place is elsewhere.

Are we good to merge now?

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -0,0 +1,67 @@
...
+You can configure the presence ping parameters in [filename]``/etc/rhn/rhn.conf``, however the default values should be sufficient in most cases.
+
+A lower total presence ping timeout value will increase the chance of false negatives.
+In some cases, a minion might be marked as non-responding, when it is responding, but did not respond quickly enough.
+A higher total presence ping timeout will increase the accuracy of the test, as even slow minions will respond to the presence ping before timing out.
+Additionally, a higher presence ping timeout could limit throughput if you are targeting a large number of minions, when some of them are slow.
+
+If a minion does not reply to a ping within the allocated time, it will be marked as [systemitem]``not available``, and will be excluded from the command.
+The {webui} will show a [systemitem]``minion is down`` message in this case.
+For more information on minion timeouts, see xref:scale-minions.adoc[].
```
moio: Totally optional suggestion: how about moving this to end of line 47?

Loquacity: That breaks the structure, which looks like this:
* Name the  parameters
* Explain the parameters
* Explain the possible effect of changing them
* List the parameter name and default

Loquacity:
> Good to go FMPOV. Thanks for the patience!

And thank you for taking the time to explain it all to me!

PR #448, opened 2019-05-01
Build errors

Loquacity:


Loquacity:
> Looks sensible. I hope an empty macro def will not cause trouble (:productname:).

Yeah, I'm not sure what's happening there. I copied across the whole entities file from the ROOT directory, when  I think I really only needed the ibmz bits. I'm going to cancel this one and try again.

Loquacity:
Please review: https://github.com/SUSE/doc-susemanager/pull/449

PR #450, opened 2019-05-02
Parallelise Troubleshooting section in Admin Guide

Loquacity:
https://github.com/SUSE/spacewalk/issues/7698


modules/administration/pages/tshoot-packages.adoc
```diff
@@ -1,23 +1,40 @@
...
+. First step
+. Another step
+. Last step
+////
+
+Packages can sometimes be locked or taskomatic can experience problems, which creates problems with metadata regeneration.
+When this occurs, package updates will be available in the {webui}, but will not appear on the client, and attempts to update the client will fail.
+To correct this, determine if any processes are running, or if a crash could have occured.
+Check package locks and exclude lists to determine if packages are locked or excluded on the client.
+Once you have located the problematic process, the metadata can be regenerated and synchronization occurs as expected.
```
keichwa:
```suggestion
When you have located the problematic process, the metadata can be regenerated and synchronization occurs as expected.
```

modules/administration/pages/tshoot-packages.adoc
```diff
@@ -1,23 +1,40 @@
...
+
+.Procedure: Resolving Package Inconsisties
+
+. On the client, check the [path]``/var/log/rhn/rhn_taskomatic_daemon.log`` configuratiopn file to determine if any processes are still running, or an exception which could indicate a crash.
```
keichwa:
```suggestion
. On the client, check the [path]``/var/log/rhn/rhn_taskomatic_daemon.log`` file to determine if any processes are still running or a crash occured.
```
the log file is not a config file ;) and the rest could be simplified, IMO.

modules/administration/pages/tshoot-registerclones.adoc
```diff
@@ -1,186 +1,128 @@
...
 
+If more detailed instructions are required, put them in a "Resolving" procedure:
+.Procedure: Resolving Widget Wobbles
+. First step
+. Another step
+. Last step
+////
 
+Sometimes a cloned client (either traditional or Salt) will use the same machine ID as the system they are a clone of.
+This results in {productname} only recognising one system, rather than two different systems.
```
keichwa:
```suggestion
This results in {productname} only recognizing one system, rather than two different systems.
```

modules/administration/pages/tshoot-registerclones.adoc
```diff
@@ -1,186 +1,128 @@
...
+This can be resolved by changing the machine ID of the  cloned system, so that {productnamne} recognises them as two different clients.
```
keichwa:
```suggestion
This can be resolved by changing the machine ID of the  cloned system, so that {productnamne} recognizes them as two different clients.
```

modules/administration/pages/tshoot-packages.adoc
mantel: And this logfile does not exist on the client, but only on the server.

modules/administration/pages/tshoot-registerclones.adoc
jcayouette: :) Nice catch Karl

modules/administration/pages/tshoot-packages.adoc
jcayouette: :+1:

modules/administration/pages/tshoot-packages.adoc
jcayouette: :+1:

PR #452, opened 2019-05-08
pxe saltboot params

keichwa:
https://github.com/SUSE/spacewalk/issues/7450

### DIFF
--- a/modules/retail/pages/retail-formulas.adoc
+++ b/modules/retail/pages/retail-formulas.adoc
@@ -246,6 +246,14 @@ Saltboot supports common kernel parameters and saltboot-specific kernel paramete
 All the parameters can be entered in the [guimenu]``Kernel Command Line Parameters`` field of the PXE formula.
 
 [systemitem]``kiwidebug=1``::
+[WARNING]
+.Do Not Use in Production
+====
+Do not use this parameter in a production environment as it creates a major security hole.
+This parameter should be used only in a development environment for debug purposes.
+====
++
+
 Starts a shell on tty2 during boot and enables debug log level in Salt.
 
 [systemitem]``MASTER``::
@@ -256,25 +264,35 @@ MASTER=myproxy.domain.com
 ----
 
 [systemitem]``SALT_TIMEOUT``::
-Fall back to the local boot if the Salt master does not apply the saltboot state within this timeout (default: 60 seconds). For example:
+Overrides the local boot fallback timeout if the Salt master does not apply the saltboot state within this timeout (default: 60 seconds).
+For example:
 +
 ----
 SALT_TIMEOUT=300
 ----
 
 // see https://bugzilla.suse.com/show_bug.cgi?id=1111348#c3
-[systemitem]``DISABLE_HOSTNAME_ID`` and [systemitem]``DISABLE_UNIQUE_SUFFIX``::
-Option for manipulating the names of the terminals.
+[systemitem]``DISABLE_HOSTNAME_ID``::
+If the terminal has a hostname assigned by DHCP, it is by default used as a minion ID.
+Setting this option to `1` disables this mechanism, and SMBios information will be used as a minion ID.
+
+[systemitem]``DISABLE_UNIQUE_SUFFIX``::
+Setting this option to `1` disables adding random generated suffix to terminal minion ID.
++
+If you set this parameter make sure your terminal has either a unique hostname provided by DHCP and DNS, or the terminal hardware comes with a unique serial number stored in its SMBios memory.
+Otherwise there is a risk of terminal minion ID duplicity, and bootstrapping the minion will fail.
+
+The following parameters (`MINION_ID_PREFIX`, `salt_device`, `root`) are usually autoconfigured and should be used only in specific conditions such as debugging or development:
 
 [systemitem]``MINION_ID_PREFIX``::
 Branch ID set in the PXE formula form.
 
-[systemitem]``salt_device``
-Device that contains the salt configuration.
+[systemitem]``salt_device``::
+Device that contains the Salt configuration.
 
-[systemitem]``root``
+[systemitem]``root``::
 Device that contains the already deployed root filesystem.
-It is used for the fallback to local boot.
+Used for falling back to local boot.
 
 
 
--- a/modules/retail/pages/retail-formulas.adoc
+++ b/modules/retail/pages/retail-formulas.adoc
@@ -273,7 +273,7 @@ Branch ID set in the PXE formula form.
 Device that contains the salt configuration.
 
 [systemitem]``root``
-device that contains the already deployed root filesystem.
+Device that contains the already deployed root filesystem.
 It is used for the fallback to local boot.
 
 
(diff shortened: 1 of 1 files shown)


modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
 
 
+
+[[retail.sect.formulas.pxe.kernelparams]]
+=== Kernel Command Line Parameters
+Saltboot supports common kernel parameters and saltboot-specific kernel parameters.
+All the parameters can be entered in the [guimenu]``Kernel Command Line Parameters`` field of the PXE formula.
+
+[systemitem]``kiwidebug=1``::
+starts a shell on tty2 during boot and enables debug log level in Salt.
```
Loquacity:
```suggestion
Starts a shell on tty2 during boot and enables debug log level in Salt.
```

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
+
+[systemitem]``MASTER``::
+overrides the Salt master autodetection; for example:
```
Loquacity:
```suggestion
Overrides auto-detection of the Salt master. For example:
```

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
++
+----
+MASTER=myproxy.domain.com
+----
+
+[systemitem]``SALT_TIMEOUT``::
+fallback to the local boot if the Salt master does not apply the saltboot state within this timeout (default: 60s); for example:
```
Loquacity:
```suggestion
Fall back to the local boot if the Salt master does not apply the saltboot state within this timeout (default: 60 seconds). For example:
```

"Fall back" is a verb, "fallback" is an adjective.

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
+
+// see https://bugzilla.suse.com/show_bug.cgi?id=1111348#c3
+[systemitem]``DISABLE_HOSTNAME_ID`` and [systemitem]``DISABLE_UNIQUE_SUFFIX``::
+Option for manipulating the names of the terminals.
+
+[systemitem]``MINION_ID_PREFIX``::
+Branch ID set in the PXE formula form.
+
+[systemitem]``salt_device``
+device that contains the salt configuration.
```
Loquacity:
```suggestion
Device that contains the salt configuration.
```

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
+
+[systemitem]``root``
+device that contains the already deployed root filesystem.
```
Loquacity:
```suggestion
Device that contains the already deployed root filesystem.
```

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
+It is used for the fallback to local boot.
```
Loquacity:
```suggestion
Used for the fall back to local boot.
```

modules/retail/pages/retail-formulas.adoc
keichwa: yes ;)  But in German we glue it together nevertheless.  But there are many other subtle differences :)

modules/retail/pages/retail-formulas.adoc
keichwa: Hmm, @Loquacity .  Maybe, we'd better say in this case:

```Used for falling back to local boot.```

File: modules/retail/pages/retail-formulas.adoc [hunk 7]
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
 
 
+
+[[retail.sect.formulas.pxe.kernelparams]]
+=== Kernel Command Line Parameters
+Saltboot supports common kernel parameters and saltboot-specific kernel parameters.
+All the parameters can be entered in the [guimenu]``Kernel Command Line Parameters`` field of the PXE formula.
+
+[systemitem]``kiwidebug=1``::
+Starts a shell on tty2 during boot and enables debug log level in Salt.
```
aaannz: Please add a warning that this should be used only in development environment and **must** not be used in production as it creates a major security hole.

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
+
+[systemitem]``MASTER``::
+Overrides auto-detection of the Salt master. For example:
++
+----
+MASTER=myproxy.domain.com
+----
+
+[systemitem]``SALT_TIMEOUT``::
+Fall back to the local boot if the Salt master does not apply the saltboot state within this timeout (default: 60 seconds). For example:
```
aaannz: Not to confuse customers, this timeout is always active and by default is 60s. This option can override this 60s default.
I would rephrase it like

`Override local boot fallback timeout if Salt master does not apply the saltboot state within this timeout (default: 60 seconds).`

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
++
+----
+SALT_TIMEOUT=300
+----
+
+// see https://bugzilla.suse.com/show_bug.cgi?id=1111348#c3
+[systemitem]``DISABLE_HOSTNAME_ID`` and [systemitem]``DISABLE_UNIQUE_SUFFIX``::
+Option for manipulating the names of the terminals.
```
aaannz: `DISABLE_UNIQUE_SUFFIX`

Setting this option to 1 disables adding random generated suffix to terminal minion id.

Please make sure your terminal has either unique hostname provided by dhcp and dns or terminal hardware comes with unique serial number stored in its SMBios memory.
Otherwise there is a risk of terminal minion id duplicity and failure to bootstrap the minion.

`DISABLE_HOSTNAME_ID`

If terminal has a hostname assigned by DHCP, it is by default used as a minion id. Setting this option to 1 disables this mechanism and SMBios information will be used as a minion id.

modules/retail/pages/retail-formulas.adoc
```diff
@@ -230,14 +230,54 @@ The PXE formula is used to configure PXE booting on the branch server.
...
+
+[systemitem]``MINION_ID_PREFIX``::
```
aaannz: Please note that these three (`MINION_ID_PREFIX`, `salt_device`, `root`) are usually autoconfigured and should be used only in specific conditions, usually during debugging/development.

modules/retail/pages/retail-formulas.adoc
keichwa: I'm going to add such a warning.

modules/retail/pages/retail-formulas.adoc
keichwa: ok

modules/retail/pages/retail-formulas.adoc
keichwa: ok

modules/retail/pages/retail-formulas.adoc
keichwa: ok

keichwa:
I think I submitted all proposed or requested changes.  Please review again, @aaannz and @Loquacity

PR #456, opened 2019-05-13
Virtualization: mention the minimum salt version

cbosdo:
Tell that the virtualization features will only work if the salt minion has 2019.2.0+ salt


cbosdo:
@jcayouette no: this is a minimum required version: it shouldn't change in the future

jcayouette:
> @jcayouette no: this is a minimum required version: it shouldn't change in the future, while the macro will change for sure.

Ah right, thanks for clarifying then leave as is :+1:

PR #457, opened 2019-05-13
Admin autoyast example channel #7740

keichwa:
enable and improve autoyast example profile.

### DIFF
--- a/modules/administration/nav-administration-guide.adoc
+++ b/modules/administration/nav-administration-guide.adoc
@@ -47,7 +47,8 @@ ifndef::backend-pdf[]
 *** xref:tshoot-rpctimeout.adoc[Troubleshooting RPC Timeouts]
 // Appendices
 // Commenting this out until we find the right place to put it. LKB 2019-03-21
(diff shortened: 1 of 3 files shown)


mbrookhuis:
Don't understand why this is added:
  <listentry>
  
    <media_url>http://$c_server/ks/dist/child/`channel-label`/`sle-manager-tools`/`distribution-label`</media_url>
    ...
  </listentry>

Think the example given and the explanation should work.
Where can I find this in the docu?

keichwa:
> Think the example given and the explanation should work.

@mbrookhuis in general, I agree.  I now added your example completely (for sles12sp4).

> Where can I find this in the docu?

We have an xref in that section where you can find info how to define variables such as `$c_server` (this xref probably needs tweaking):
```
For information about variables, see
ifdef::env-github,backend-html5[]
<<reference-webui-systems-autoinstallation.adoc#s4-sm-system-kick-dist-variables, Autoinstallable Distribution Variables>>.
endif::[]
ifndef::env-github,backend-html5[]
<<s4-sm-system-kick-dist-variables>>.
endif::[]
```

PR #469, opened 2019-05-19
Use 'My' instead of 'Your' for menu and title pages

ncounter:
See https://github.com/uyuni-project/uyuni/pull/988

Open question: do files need to be renamed as well or we can keep them as they are right now? For instance, I am thinking  about [this file](https://github.com/SUSE/doc-susemanager/blob/master/modules/reference/pages/home/home-your-preferences.adoc) actually.


Loquacity:
> Open question: do files need to be renamed as well or we can keep them as they are right now? For instance, I am thinking about [this file](https://github.com/SUSE/doc-susemanager/blob/master/modules/reference/pages/home/home-your-preferences.adoc) actually.

Yeah, that's probably a good idea.

PR #474, opened 2019-05-23
Disable Salt Mine

Loquacity:
https://github.com/SUSE/spacewalk/issues/7731

### DIFF
--- a/modules/salt/pages/salt-rate-limiting.adoc
+++ b/modules/salt/pages/salt-rate-limiting.adoc
@@ -66,3 +66,26 @@ Defaults to 4 seconds.
 The presence ping gather job parameter changes the timeout setting for gathering the presence ping, in seconds.
 Adjust the [systemitem]``java.salt_presence_ping_gather_job_timeout`` parameter.
 Defaults to 1 second.
+
+
+== Disabling the Salt Mine
+
+In older versions, {productname} used a tool called Salt Mine to check minion availability.
+With the introduction of more efficient tools in {productname}{nbsp}3.2, it is no longer required.
+If the Salt mine is running on your system, you can reduce load by disabling it.
+Disable the Salt mine by running this command on the server:
+
+----
+salt '*' state.sls util.mgr_mine_config_clean_up
+----
+
+This will restart the minions and generate some Salt events to be processed by the server.
+However, if you have a large number of minions, handling these events could create excessive load.
+To avoid this, you can execute the command in batch mode with this command:
+
+----
+salt --batch-size 50 '*' state.sls util.mgr_mine_config_clean_up
+----
+
+You will need to wait for this command to finish executing.
+Do not end the process with kbd:[Ctrl+C].
--- a/modules/salt/pages/salt-rate-limiting.adoc
+++ b/modules/salt/pages/salt-rate-limiting.adoc
@@ -71,9 +71,9 @@ Defaults to 1 second.
 == Disabling the Salt Mine
 
 In older versions, {productname} used a tool called Salt mine to check minion availability.
-Salt mine would ping the server every hour, which created significant load.
-With the introduction of more efficient tools in {productname}{nbsp}3.2, Salt mine is no longer required.
-Instead, the {productname} server uses Taskomatic to ping only the minions that appear to have been offline for twelve hours or more, with all minions being pinged only once in every twenty four hour period by default.
+Salt mine would cause minions to ping the server every hour, which created significant load.
+With the introduction of a more efficient mechanism in {productname}{nbsp}3.2, Salt mine is no longer required.
+Instead, the {productname} server uses Taskomatic to ping only the minions that appear to have been offline for twelve hours or more, with all minions being pinged at least once in every twenty four hour period by default.
 
 Newly registered Salt minions will have the Salt mine disabled by default.
 If the Salt mine is running on your system, you can reduce load by disabling it.
--- a/modules/salt/pages/salt-rate-limiting.adoc
+++ b/modules/salt/pages/salt-rate-limiting.adoc
@@ -71,8 +71,14 @@ Defaults to 1 second.
 == Disabling the Salt Mine
 
 In older versions, {productname} used a tool called Salt mine to check minion availability.
-With the introduction of more efficient tools in {productname}{nbsp}3.2, it is no longer required.
+Salt mine would ping the server every hour, which created significant load.
+With the introduction of more efficient tools in {productname}{nbsp}3.2, Salt mine is no longer required.
+Instead, the {productname} server uses Taskomatic to ping only the minions that appear to have been offline for twelve hours or more, with all minions being pinged only once in every twenty four hour period by default.
+
+Newly registered Salt minions will have the Salt mine disabled by default.
 If the Salt mine is running on your system, you can reduce load by disabling it.
+This is especially effective if you have a large number of minions.
+
 Disable the Salt mine by running this command on the server:
 
 ----
--- a/modules/salt/pages/salt-rate-limiting.adoc
+++ b/modules/salt/pages/salt-rate-limiting.adoc
@@ -70,7 +70,7 @@ Defaults to 1 second.
 
 == Disabling the Salt Mine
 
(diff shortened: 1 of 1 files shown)


modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -66,3 +66,26 @@ Defaults to 4 seconds.
 The presence ping gather job parameter changes the timeout setting for gathering the presence ping, in seconds.
 Adjust the [systemitem]``java.salt_presence_ping_gather_job_timeout`` parameter.
 Defaults to 1 second.
+
+
+== Disabling the Salt Mine
+
+In older versions, {productname} used a tool called Salt mine to check minion availability.
```
chiaradiamarcelo: We could add how much load we generated before, so somehow we explain why the new way is better. Something like:
"In older versions, {productname} used a tool called Salt mine to check minion availability, pinging the server every hour."

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -66,3 +66,26 @@ Defaults to 4 seconds.
...
+With the introduction of more efficient tools in {productname}{nbsp}3.2, it is no longer required.
```
chiaradiamarcelo: Maybe here we could talk a bit about the new mechanism, something like: " now the Server has a Taskomatic task that will ping only those minions that appear offline since at least 12 hours, with the target of pinging all of them at least once every 24 hours (by default)." (that's actually how it works, so that's why the events load is much smaller)

chiaradiamarcelo: We can also say that the newly registered minions does not have the Salt mine enabled by default anymore (since 3.2.6), but for the already registered is recommended to disabled it, especially if the minion count is big.

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -71,8 +71,14 @@ Defaults to 1 second.
 == Disabling the Salt Mine
 
 In older versions, {productname} used a tool called Salt mine to check minion availability.
-With the introduction of more efficient tools in {productname}{nbsp}3.2, it is no longer required.
+Salt mine would ping the server every hour, which created significant load.
+With the introduction of more efficient tools in {productname}{nbsp}3.2, Salt mine is no longer required.
+Instead, the {productname} server uses Taskomatic to ping only the minions that appear to have been offline for twelve hours or more, with all minions being pinged only once in every twenty four hour period by default.
```
chiaradiamarcelo: "with all minions being pinged only once in every twenty four hour period by default." -> "with all minions being pinged at least once in every twenty four hour period by default"

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -71,8 +71,14 @@ Defaults to 1 second.
 == Disabling the Salt Mine
 
 In older versions, {productname} used a tool called Salt mine to check minion availability.
-With the introduction of more efficient tools in {productname}{nbsp}3.2, it is no longer required.
+Salt mine would ping the server every hour, which created significant load.
+With the introduction of more efficient tools in {productname}{nbsp}3.2, Salt mine is no longer required.
```
chiaradiamarcelo: maybe we could use "With the introduction of a more efficient mechanism" instead of "more efficient tools"

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -66,3 +66,32 @@ Defaults to 4 seconds.
 The presence ping gather job parameter changes the timeout setting for gathering the presence ping, in seconds.
 Adjust the [systemitem]``java.salt_presence_ping_gather_job_timeout`` parameter.
 Defaults to 1 second.
+
+
+== Disabling the Salt Mine
+
+In older versions, {productname} used a tool called Salt mine to check minion availability.
+Salt mine would ping the server every hour, which created significant load.
```
chiaradiamarcelo: Hmm this is not actually accurate. What happens actually is, when the Salt mine is enabled on the minions, an event is triggered every hour from the minion and is processed by the Server, and we use that as a "checkIn".

Maybe we could rephrase to something like "With the Salt mine enabled, the minions would checkIn every hour".

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -71,9 +71,9 @@ Defaults to 1 second.
 == Disabling the Salt Mine
 
 In older versions, {productname} used a tool called Salt mine to check minion availability.
-Salt mine would ping the server every hour, which created significant load.
-With the introduction of more efficient tools in {productname}{nbsp}3.2, Salt mine is no longer required.
-Instead, the {productname} server uses Taskomatic to ping only the minions that appear to have been offline for twelve hours or more, with all minions being pinged only once in every twenty four hour period by default.
+Salt mine would cause minions to ping the server every hour, which created significant load.
```
chiaradiamarcelo: "The Salt mine"

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -71,9 +71,9 @@ Defaults to 1 second.
...
+With the introduction of a more efficient mechanism in {productname}{nbsp}3.2, Salt mine is no longer required.
```
chiaradiamarcelo: "the Salt mine is no longer required..."

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -66,3 +66,32 @@ Defaults to 4 seconds.
 The presence ping gather job parameter changes the timeout setting for gathering the presence ping, in seconds.
 Adjust the [systemitem]``java.salt_presence_ping_gather_job_timeout`` parameter.
 Defaults to 1 second.
+
+
+== Disabling the Salt Mine
+
+In older versions, {productname} used a tool called Salt mine to check minion availability.
+The Salt mine would cause minions to ping the server every hour, which created significant load.
```
moio: Technically this is not "pinging", I would call this just "contacting".

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -66,3 +66,32 @@ Defaults to 4 seconds.
...
+With the introduction of a more efficient mechanism in {productname}{nbsp}3.2, the Salt mine is no longer required.
+Instead, the {productname} server uses Taskomatic to ping only the minions that appear to have been offline for twelve hours or more, with all minions being pinged at least once in every twenty four hour period by default.
```
moio: Do we want to describe how to change that default, or link it from here?

modules/salt/pages/salt-rate-limiting.adoc
```diff
@@ -66,3 +66,32 @@ Defaults to 4 seconds.
...
+This is especially effective if you have a large number of minions.
+
+Disable the Salt mine by running this command on the server:
+
+----
+salt '*' state.sls util.mgr_mine_config_clean_up
+----
+
+This will restart the minions and generate some Salt events to be processed by the server.
+However, if you have a large number of minions, handling these events could create excessive load.
```
moio: I would use "Thus" instead of "However" here.

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: I can't immediately see where we've documented this, so maybe we should write it up.

modules/salt/pages/salt-rate-limiting.adoc
Loquacity: Took it out  entirely.

Loquacity:
@moio Please review https://github.com/SUSE/doc-susemanager/pull/492

modules/salt/pages/salt-rate-limiting.adoc
moio: It's the `web.system_checkin_threshold` parameter in `rhn.conf`, expressed in days, default value is 1.

Loquacity:
> Late review, sorry. Please also note this is also to be backported to 3.2.

Please review: https://github.com/SUSE/doc-susemanager/pull/499

PR #477, opened 2019-05-23
Proxy intro wording ke

keichwa:
change wording in intro


keichwa:
does not work as expected :-(((

PR #485, opened 2019-05-27
Minor edits to the new Prometheus section

Loquacity:
Minor edits to clean up after the merge on Friday.

### DIFF
--- a/modules/administration/pages/prometheus.adoc
+++ b/modules/administration/pages/prometheus.adoc
@@ -115,11 +115,15 @@ See https://github.com/boynux/squid-exporter.
 
 
 
-== Prometheus Configuration Steps
+== Install and Configure Prometheus
 
-=== Prometheus Installation
+Prometheus is installed from a package, and needs configuration before you can use it to gather metrics.
 
-.Procedure: Setup Prometheus
+
+
+=== Installing Prometheus
+
+.Procedure: Installing Prometheus
 . Install the [package]``golang-github-prometheus-prometheus`` package:
 +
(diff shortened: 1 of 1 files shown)


modules/administration/pages/prometheus.adoc
```diff
@@ -130,7 +134,8 @@ zypper in golang-github-prometheus-prometheus
 ----
 systemctl enable --now prometheus
 ----
-. Navigate to port 9090 in your browser and confirm the Prometheus UI is loaded.
+. Navigate to port 9090 in your browser to confirm that  the Prometheus interface is loading correctly.
```
keichwa: Sounds strange.  I guess it is meant to open the Prometheus page through port 9090. Or use a tool such as telnet for checking the port.

cavalheiro: I'd say we can recommend the browser instead of using telnet, as users can try navigating through the UI and check if everything is in place.

keichwa: Oops, sorry, my statement was not clear enough.  I'm fine with using the browser, but I would not know how to navigate to port 9090 in the browser (maybe, I lack context, I just see the diff and I'm not familiar with P).

cavalheiro: What if we write something like "On your browser, navigate to the URL of the server where you installed Prometheus, on port 9090 (ex: http://my-prometheus-server.com:9090) and confirm that the Prometheus interface is loading correctly." ?

PR #520, opened 2019-06-06
Add admonition about SSL certs trust.

Loquacity:
https://bugzilla.suse.com/show_bug.cgi?id=1136561
https://github.com/SUSE/spacewalk/issues/8023

### DIFF
--- a/modules/administration/pages/virtual-hosts.adoc
+++ b/modules/administration/pages/virtual-hosts.adoc
@@ -111,4 +111,10 @@ If you do not refresh the data from a new Virtual Host Manager, host data will n
 ====
 +
 
-. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems] .
+. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems].
+
+[NOTE]
+====
+Depending on your environment and settings, you might get an ``invalid certificate`` error when you connect to the server.
+If this occurs, check that your SSL certificate is current, run the [command]``update-ca-certificates`` command on the command line to re-trust your certificate, and restart the spacewalk services.
+====
--- a/modules/administration/pages/virtual-hosts.adoc
+++ b/modules/administration/pages/virtual-hosts.adoc
@@ -115,6 +115,6 @@ If you do not refresh the data from a new Virtual Host Manager, host data will n
 
 [NOTE]
 ====
-Depending on your environment and settings, you may recieve an ``invalid certificate`` error when you connect to the server.
+Depending on your environment and settings, you might receive an ``invalid certificate`` error when you connect to the server.
 If this occurs, check that your SSL certificate is current, run the [command]``update-ca-certificates`` command on the command line to re-trust your certificate, and restart the spacewalk services.
 ====
--- a/modules/administration/pages/virtual-hosts.adoc
+++ b/modules/administration/pages/virtual-hosts.adoc
@@ -111,7 +111,7 @@ If you do not refresh the data from a new Virtual Host Manager, host data will n
 ====
 +
 
-. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems].
+. View inventoried ESXi host objects and resources by selecting menu:System List[Virtual Systems].
 
 [NOTE]
 ====
--- a/modules/administration/pages/virtual-hosts.adoc
+++ b/modules/administration/pages/virtual-hosts.adoc
@@ -9,7 +9,7 @@
 
 Foreign virtual hosts (such as vCenter and vSphere ESXi) can be inventoried using the [guimenu]``Virtual Host Manager``.
 From the vSphere Client you can define roles and permissions for vCenter and vSphere ESXi users allowing vSphere objects and resources to be imported and inventoried by {productname}.
-Objects and resources are then displayed as foreign hosts on the {productname} menu:Systems[Virtual Systems] page.
+Objects and resources are then displayed as foreign hosts on the {productname} menu:System List[Virtual Systems] page.
 
 The following sections will guide you through:
 
@@ -107,7 +107,7 @@ image::systems_virtual_host_managers_vmware.png[scaledwidth=80%]
 
 [IMPORTANT]
 ====
-If you do not refresh the data from a new Virtual Host Manager, host data will not be inventoried and therefore will not be displayed under menu:Systems[Virtual Systems].
+If you do not refresh the data from a new Virtual Host Manager, host data will not be inventoried and therefore will not be displayed under menu:System List[Virtual Systems].
 ====
 +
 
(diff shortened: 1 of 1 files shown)


Loquacity:
> LGTM.

I removed 'may'. 'May' has the primary meaning of "have permission to" (in addition to the secondary "a possibility of", the meaning you're using here), which in computing terms can get confusing. I tend to avoid "may", preferring either "can" or "might". Modal verbs are confusing, though, and everyone has a personal preference. I also just tend to think that 'may' sounds pretentious :stuck_out_tongue_closed_eyes: 

Other than that, you typoed on 'receive'.

modules/administration/pages/virtual-hosts.adoc
```diff
@@ -111,4 +111,10 @@ If you do not refresh the data from a new Virtual Host Manager, host data will n
 ====
 +
 
-. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems] .
+. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems].
```
keichwa: You changes the second level "Systems" to "System List" and you did not tell us. :unamused: Thus we not either must use as a short version:

```suggestion
. View inventoried ESXi host objects and resources by selecting menu:System List[Virtual Systems].
```

Or the long version:
```
menu:Main Menu[Systems > System List > Virtual Systems]
```
But do not worry too much. There are inconsistencies with marking up menu entries everywhere...  And I do not know what we can do about it.

modules/administration/pages/virtual-hosts.adoc
```diff
@@ -111,4 +111,10 @@ If you do not refresh the data from a new Virtual Host Manager, host data will n
...
+
+[NOTE]
+====
+Depending on your environment and settings, you might receive an ``invalid certificate`` error when you connect to the server.
+If this occurs, check that your SSL certificate is current, run the [command]``update-ca-certificates`` command on the command line to re-trust your certificate, and restart the spacewalk services.
```
Bischoff: I know that @mcalmer suggested to avoid precise instructions here. Still, I would prefer something more explicit:
```
check that your SSL certificate is current
```
=>
```
extract the certificate from the ESXi server and copy it into `/etc/pki/trust/anchors`
```

Bischoff: Also, I tested, and it worked for me *without* restarting the spacewalk services.

modules/administration/pages/virtual-hosts.adoc
```diff
@@ -111,4 +111,10 @@ If you do not refresh the data from a new Virtual Host Manager, host data will n
 ====
 +
 
-. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems] .
+. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems].
+
+[NOTE]
+====
+Depending on your environment and settings, you might receive an ``invalid certificate`` error when you connect to the server.
```
Bischoff: I suggest being more explicit:
```
when you connect to the ESXi server via HTTPS protocol from a web browser.
If that's the case, refreshing the data from the virtual hosts server will not work as expected.
```

modules/administration/pages/virtual-hosts.adoc
Loquacity: > 
> Or this long version:
> 
> ```
> menu:Main Menu[Systems > System List > Virtual Systems]
> ```
> 
I dislike the "main menu" thing, it just makes it laborious (to write and to read).

> But do not worry too much. There are inconsistencies with marking up menu entries everywhere... And I do not know what we can do about it.

Do a consistency sweep and update them all? We need to do this for a few things, I think ... (post 4.0 jobs!)

modules/administration/pages/virtual-hosts.adoc
Loquacity: I think it's probably best to leave it in. If they do it and don't need to, then no harm done.

modules/administration/pages/virtual-hosts.adoc
```diff
@@ -107,8 +107,16 @@ image::systems_virtual_host_managers_vmware.png[scaledwidth=80%]
...
 ====
 +
 
-. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems] .
+. View inventoried ESXi host objects and resources by selecting menu:System List[Virtual Systems].
+
+[NOTE]
+====
+Depending on your environment and settings, you might receive an ``invalid certificate`` error when you connect to the ESXi server from a web browser using HTTPS.
+If this occurs, refreshing the data from the virtual hosts server will fail.
```
Bischoff: We could simply remove "Depending on your environment and settings, ", because it does not bring much information and is not even really true.

To be very honest, it will not "fail" with a message, but silently provide no result... The failure can be seen only in the logs. But it's a product problem, not a documentation issue. So unrelated to this PR.

modules/administration/pages/virtual-hosts.adoc
```diff
@@ -107,8 +107,16 @@ image::systems_virtual_host_managers_vmware.png[scaledwidth=80%]
...
+If you do not refresh the data from a new Virtual Host Manager, host data will not be inventoried and therefore will not be displayed under menu:System List[Virtual Systems].
 ====
 +
 
-. View inventoried ESXi host objects and resources by selecting menu:Systems[Virtual Systems] .
+. View inventoried ESXi host objects and resources by selecting menu:System List[Virtual Systems].
+
+[NOTE]
+====
+Depending on your environment and settings, you might receive an ``invalid certificate`` error when you connect to the ESXi server from a web browser using HTTPS.
```
keichwa:
```suggestion
Depending on your environment and settings, you might receive an ``invalid certificate`` error when you connect to the ESXi server from a Web browser using HTTPS.
```
According to the style guide, we used to capitalize Web.

Loquacity: Seems old fashioned ... Do we capitalise Internet as well?

Loquacity: Actually, on second thoughts, we can probably just remove the word 'web' ... which removes the problem ;)

keichwa:

> Actually, on second thoughts, we can probably just remove the word 'web'
> ... which removes the problem ;)

:+1:  Yes, I think we also capitalize Internet ;)  It's after
2000-policy, though.  (I admit, I did not check the suse style guide
lately.)


-- 
                           90409 Nürnberg, Germany

Bischoff:
> Actually, on second thoughts, we can probably just remove the word 'web' ... which removes the problem ;)
> +1 Yes, I think we also capitalize Internet ;) It's post-2000 policy, though. (I admit, I did not check the suse style guide lately.)

@Loquacity In my tech writing company, we adopted the convention to capitalize (and possibly use an article) for  Internet and Web used as nouns, and no capital when used as an adjective.

I'm not sure how official this quite common convention is though.

Loquacity:
> 
> @keichwa @Loquacity In my tech writing company, we adopted the convention to capitalize (and possibly use an article) for Internet and Web used as nouns, and no capital when used as adjectives.
> 
> I'm not sure how official this quite common convention is, though...

The problem with language is it doesn't stand still. This was certainly 'standard' in the 00s, but is considered a little anachronistic these days.

PR #527, opened 2019-06-11
Retail formula chapter split

Loquacity:
Splits formulas chapter into sections.


modules/retail/nav-retail.adoc
```diff
@@ -10,9 +10,18 @@ ifndef::backend-pdf[]
 *** xref:retail-install-packages.adoc[Installing from Packages]
 *** xref:retail-install-setup.adoc[Setting up]
 *** xref:retail-deploy-terminals.adoc[Deploying Terminals]
+// Formulas
+** xref:retail-formulas.adoc[Formulas]
```
keichwa: This diff tells me that you deleted `retail-formulas.adoc`, but here is ist still referenced.

keichwa: otherwise fine.  assuming that there are no additional wording changes (I did not read all the text).

Loquacity: Good spot! Fixed now :)

I haven't changed any text.

PR #528, opened 2019-06-11
Autoinstall Intro rewrite

Loquacity:
Question: What is the alternative to AutoYast/Kickstart for Salt minions? @moio?

### DIFF
--- a/modules/client-configuration/pages/autoinstallation-methods.adoc
+++ b/modules/client-configuration/pages/autoinstallation-methods.adoc
@@ -2,28 +2,16 @@
 = Introduction
 
 
+Kickstart and {ay} configuration files allow you to automate system installations for traditional clients.
+This is especially useful if you need to install a large number of clients.
 
+For {sle} clients, use {ay}.
+When you have created an {ay} file, you can upload it and manage it with {productname}.
 
-.Autoinstallation Types: {ay} and Kickstart
-[NOTE]
-====
-In the following section, {ay} and {ay} features apply for {sle} client systems only.
+For {rhel} clients, use Kickstart.
+Kickstart files are created, modified, and managed within the {productname} {webui}.
 
-For {rhela} systems, use Kickstart and Kickstart features.
-====
+The Cobbler installation server allows you to automate baremetal installations.
+It uses DHCP to access a PXE boot server, and can be used in virtualized environments.
 
-//TODO We need to add an alternative explanation for automating salt installations. This is what Salt was designed for. We should provide examples to clarify. The following admon comes acrross as "negative".
-
-.Auto-Installing Salt Minions Currently Not Supported
-[IMPORTANT]
-====
-This procedure will work for traditionally  managed systems (system type [systemitem]``management``).
-
-Autoinstallation is not currently available for systems using Salt (system type [systemitem]``salt``).
-====
-
-{ay} and Kickstart configuration files allow administrators to create an environment for automating otherwise time-consuming system installations, such as multiple servers or workstations. {ay} files have to be uploaded to be managed with {productname}.
-Kickstart files can be created, modified, and managed within the {productname} Web interface.
-
-{productname} also features the Cobbler installation server.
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/autoinstallation-methods.adoc
```diff
@@ -2,28 +2,16 @@
...
-.Autoinstallation Types: {ay} and Kickstart
-[NOTE]
-====
-In the following section, {ay} and {ay} features apply for {sle} client systems only.
+For {rhel} clients, use Kickstart.
+Kickstart files are created, modified, and managed within the {productname} {webui}.
 
-For {rhela} systems, use Kickstart and Kickstart features.
-====
+The Cobbler installation server allows you to automate baremetal installations.
```
keichwa:
```suggestion
The Cobbler installation server allows you to automate bare-metal installations.
```
(we spell it this way in the Web UI: Admin > Manager Config > Bare-metal Systems.)  In general, the term seems to be deprecated these days.  I think we should replace it with "physical system installations", or something similar.

moio:
AutoYAST and Kickstart live outside of the client management stack, so they will work regardless.

We do not have Salt integration with those two projects in SUSE Manager/Uyuni though - at the moment the only way is still to use traditional clients.

modules/client-configuration/pages/autoinstallation-methods.adoc
Loquacity: We probably should make it bare metal, according to this: https://books.google.com/ngrams/graph?content=baremetal%2Cbare-metal%2Cbare+metal&year_start=1900&year_end=2008&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cbaremetal%3B%2Cc0%3B.t1%3B%2Cbare%20-%20metal%3B%2Cc0%3B.t1%3B%2Cbare%20metal%3B%2Cc0

Although I agree with you on using a different term altogether, that's probably a different argument :wink:

Loquacity:
> AutoYAST and Kickstart live outside of the client management stack, so they will work regardless.
> 
> We do not have Salt integration with those two projects in SUSE Manager/Uyuni though - at the moment the only way is still to use traditional clients.

I don't understand the semantic difference here, sorry :(

moio:
AutoYAST is an open source project that stands on its own - it's part of all SUSE distros and it allows one to install a distro non-interactively. You write a configuration file and pass it to the kernel at boot time, and that configuration time describes all things you would normally specify during the distro installation: timezone, users, passwords, package selection and so on.

It's basically a component to automate installations, so that a user that has to install, say, 1000 machines does not have to physically sit in front of each going through the installer screens.


Kickstart is the equivalent for the Red world (and also works on Ubuntu to some degree).


Of course, since SUSE Manager/Uyuni are "server lifecycle tools" it makes sense for them to support AutoYAST and Kickstart, as installation is typically the first step in the management of a machine's lifecycle.

Such integration is only implemented for the traditional stack.

Loquacity:
> AutoYAST is an open source project that stands on its own - it's part of all SUSE distros and it allows one to install a distro non-interactively. You write a configuration file and pass it to the kernel at boot time, and that configuration time describes all things you would normally specify during the distro installation: timezone, users, passwords, package selection and so on.
> 
> It's basically a component to automate installations, so that a user that has to install, say, 1000 machines does not have to physically sit in front of each going through the installer screens.
> 
> Kickstart is the equivalent for the Red world (and also works on Ubuntu to some degree).

Yep, that was my understanding.

> 
> Of course, since SUSE Manager/Uyuni are "server lifecycle tools" it makes sense for them to support AutoYAST and Kickstart, as installation is typically the first step in the management of a machine's lifecycle.

Agreed.

> 
> Such integration is only implemented for the traditional stack.

OK, so my question still stands, then: do we have an equivalent feature for Salt?

Loquacity:
After a good conversation with @moio we've determined that the distinction about traditional/salt clients is important, it's also slightly more complicated, and doesn't need to be explained on this page. This PR should be ready to merge now.

PR #531, opened 2019-06-12
Remove Grafana from Monitoring Chapter

mcalmer:
As long as we do not know how we can ship it, we should not have it in the docs

### DIFF
--- a/modules/administration/pages/prometheus.adoc
+++ b/modules/administration/pages/prometheus.adoc
@@ -9,12 +9,10 @@ Prometheus is a monitoring tool, originally built at SoundCloud, that is used to
 Unlike other monitoring systems, Prometheus collects metrics using HTTP pulls, allowing for higher performance and scalability.
 Prometheus is an open-source software project, mostly written in Go, and its source code is available at https://github.com/prometheus/.
 
-////
 Grafana is a tool for data visualization, monitoring and analysis.
 It is used to create dashboards with panels representing specific metrics over a set period of time.
 Grafana is commonly used together with Prometheus, but also supports other data sources such as ElasticSearch, MySQL, PostgreSQL, and Influx DB.
 For more information about Grafana, see: https://grafana.com/docs/.
-////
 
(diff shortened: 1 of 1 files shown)


mcalmer:
@Loquacity - we are still discussing if we need to remove it. This is just to prepare for it.

mcalmer:
The text was discussed with @joesusecom . It is not nice but should solve our problems util we have a better soltion.
@Loquacity @jcayouette - we need to merge it and we need a new package for GMC4 and new content deployed on the web.

modules/administration/pages/prometheus.adoc
```diff
@@ -14,7 +14,7 @@ It is used to create dashboards with panels representing specific metrics over a
 Grafana is commonly used together with Prometheus, but also supports other data sources such as ElasticSearch, MySQL, PostgreSQL, and Influx DB.
```
cavalheiro: Since we are removing all the references to Grafana, should we keep this paragraph and the one above, where we present what Grafana is?

In case we still want to present Grafana here - although not shipping the package - maybe we could also include the link to the SUSE Manager dashboard on grafana.com (below, line ~216), since users that manually download and install Grafana might still benefit from this.

modules/administration/pages/prometheus.adoc
```diff
@@ -210,8 +210,7 @@ image::monitoring_enable_services.png[scaledwidth=80%]
 image::monitoring_configure_formula.png[scaledwidth=80%]
 
 
-
-== Set up Visualization with Grafana
+== Visualization with Grafana
 
 The Grafana website contains dozens of dashboards uploaded by the community.
```
cavalheiro: Maybe keep this?

mcalmer: I just removed the "set up" and the topic about installing the package

modules/administration/pages/prometheus.adoc
mcalmer: with my 2nd commit this is back again

PR #536, opened 2019-06-14
Added SLES for SAP

Loquacity:
https://bugzilla.suse.com/show_bug.cgi?id=1137898
https://github.com/SUSE/spacewalk/issues/8113

### DIFF
--- a/modules/installation/pages/client-requirements.adoc
+++ b/modules/installation/pages/client-requirements.adoc
@@ -23,15 +23,13 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
 | {sle} 11 | {x86}, {x86}_64, {ipf}, IBM {ppc}, {ibmz}  | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 7 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 6 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
-| {rhel} 8 | {x86}_64                                   | icon:question[role="gray"] | icon:question[role="gray"]
+| {sls} for SAP | {x86}_64                              | icon:check[role="green"]      | icon:check[role="green"]
+| {rhel} 8 | {x86}_64                                   | icon:question[role="gray"]    | icon:question[role="gray"]
 | {rhel} 7 | {x86}_64                                   | icon:check[role="green"]      | icon:check[role="green"]
 | {rhel} 6 | {x86}, {x86}_64                            | icon:check[role="green"]      | icon:check[role="green"]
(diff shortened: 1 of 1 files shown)


modules/installation/pages/client-requirements.adoc
```diff
@@ -23,15 +23,13 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
 | {sle} 11 | {x86}, {x86}_64, {ipf}, IBM {ppc}, {ibmz}  | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 7 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 6 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
-| {rhel} 8 | {x86}_64                                   | icon:question[role="gray"] | icon:question[role="gray"]
+| {sls} for SAP | {x86}_64                              | icon:check[role="green"]      | icon:check[role="green"]
```
mcalmer: I see a lot of follow up questions comming...

We need to add versions similar to what we say for normal SLES and we need to check architectures.
For 12 and 15 we also support ppc64le as well. Not sure about version 11.

Loquacity: Seems excessive to me, honestly. How deep does this rabbit hole go?

mcalmer: Either we list all versions and architectures or we say explicit 'all' versions and 'all' architectures.
Otherwise we will get back questions. We tried already to list only some products and everytime we got bugs to add also this and that.

Loquacity: *sigh*

OK, tell me what I need to add?

mcalmer: Hmm, with all these macros used I lost a bit the idea how the whole table look like.

Now I had a look at the "resulting" page and saw that we talk about 

  "SUSE Linux Enterprise <VERSION>" 

which is pretty generic and include all kind of products. 
When we start to talk about "for SAP" specially we need to say also "Server", "Desktop", "HPC" and whatever comes next.

What about a footnote saying it includes the following products in the available architectures ... 
and list them there.

But now my feeling is we should discuss it also with PM.

Loquacity: I've emailed Benajmin Landorf, @joesusecom, @paususe, and @mcalmer for advice.

Loquacity: Bumped email 2019-07-01

Loquacity: I haven't gotten any response to this at all. I'm going to merge and close the ticket and then we can see who yells :woman_shrugging:

mcalmer: But please fix the architecture. SLES for SAP is also available for ppc64le.

Loquacity:
@mcalmer you're blocking my merge, can you please review?

PR #538, opened 2019-06-14
Port 8050 needs to be opened for graphical console display

cbosdo:
The VM Graphical Console feature requires the port 8050 to be reachable from the browser. Document this.


Bischoff:
And what about the need for a certificate ?

cbosdo:
This is the same certificate than SUSE Manager's HTTPS port, shouldn't require additional documentation, unless customer is in test environment with self-signed certificate.

Bischoff:
ok.

Loquacity:
This will also need to be listed in modules/installation/pages/ports.adoc

PR #539, opened 2019-06-14
4.0 - Document 8050 port opening

cbosdo:
The VM Graphical Console feature requires the port 8050 to be reachable from the browser. Document this.

Port of #538.


Loquacity:
This will also need to be listed in modules/installation/pages/ports.adoc

PR #535, opened 2019-06-13
Add a section about 'Using Content Lifecycle Project'

hustodemon:
... +  minor fixes of the existing sections.

Review commit-by-commit please.

PR #540, opened 2019-06-17
Edits to CLM section

Loquacity:
Follow up from https://github.com/SUSE/doc-susemanager/pull/535

Related to https://bugzilla.suse.com/show_bug.cgi?id=1137955 and https://github.com/SUSE/spacewalk/issues/8093

### DIFF
--- a/modules/administration/pages/content-lifecycle.adoc
+++ b/modules/administration/pages/content-lifecycle.adoc
@@ -43,7 +43,7 @@ Content filters are global in your organization and can be shared between projec
 ====
 If your project already contains built sources, when you add an environment it will automatically populate with the existing content.
 Content will be drawn from the previous environment of the cycle if it had one.
-If there is no previous environment, it will use the latest available built version.
+If there is no previous environment, it will be left empty until the project sources are built again.
 ====
 
 
--- a/modules/administration/pages/content-lifecycle.adoc
+++ b/modules/administration/pages/content-lifecycle.adoc
@@ -42,6 +42,8 @@ Content filters are global in your organization and can be shared between projec
 [NOTE]
 ====
 If your project already contains built sources, when you add an environment it will automatically populate with the existing content.
+Content will be drawn from the previous environment of the cycle if it had one.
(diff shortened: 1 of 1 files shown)


modules/administration/pages/content-lifecycle.adoc
```diff
@@ -37,49 +35,54 @@ The [guimenu]``Name`` field only accepts lowercase letters, numbers, periods (``
...
 ====
-Content Filters are global in your organization and can be shared in multiple projects.
+Content filters are global in your organization and can be shared between projects.
 ====
 
 [NOTE]
-.Adding environment in existing project
 ====
-When you add an environment to an existing project that has already some sources built, the newly added environment will be populated with the content of its predecessor environment.
+If your project already contains built sources, when you add an environment it will automatically populate with the existing content.
```
ncounter: I think the "predecessor" part is important to be mentioned. I like the rephrase but we should still say that. Maybe something like:
`If your project already contains built sources, when you add an environment it will automatically populate with the existing content from the previous environment of the cycle or, if the new environment takes the first place, it will get the latest sources built version.`

modules/administration/pages/content-lifecycle.adoc
```diff
@@ -37,49 +35,56 @@ The [guimenu]``Name`` field only accepts lowercase letters, numbers, periods (``
...
+Content will be drawn from the previous environment of the cycle if it had one.
+If there is no previous environment, it will use the latest available built version.
```
ncounter: And I just realized this "part of the feature" is not yet implemented at the moment. It will be, but not right now. We need to remove the case when `there is no previous environment`, or we need to clarify rewording it in `If there is no previous environment, it will be left empty, waiting for a new Build of the project sources."

Loquacity:
> I am so happy you overtake my feedback and you do a proper rewording of the content. I thought I wrote a good English sentence actually, but then......I read your version and yes, **that on is** a documentation paragraph. heavy_plus_sign 100

Aww, thanks for the kind words. All those years of school taught me something!

PR #551, opened 2019-07-02
Add Tuning info about changelog metadata storage.

Loquacity:
https://github.com/SUSE/spacewalk/issues/7090

### DIFF
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -1,0 +1,16 @@
+[[tuning-changelogs]]
+= Tuning Changelogs
+
+Some packages have a long list of changelog entries.
+This data is downloaded by default, but is very rarely useful information.
+In order to limit the amount of changelog metadata which is downloaded and to save disk space, you can put a limit on how many entries to keep on disk.
+
+This configuration option is in the [filename]``/etc/rhn/rhn.conf`` configuration file.
+The parameter defaults to [systemitem]``0``, which means unlimited.
+
+----
+java.max_changelog_entries = 0
+----
+
+When you set this parameter, it will come into effect only when packages are synchronized.
+You might like to delete the cached data to remove older data.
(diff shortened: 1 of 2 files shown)


modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -0,0 +1,16 @@
+[[tuning-changelogs]]
+= Tuning Changelogs
+
+Some packages have a long list of changelog entries.
+This data is downloaded by default, but is very rarely useful information.
```
mcalmer: We can keep this even when it is not really correct. changelog info are useful. But most of the tools used for installing updates cannot show them. They download data which are not used.

In addition the question is, if 10 years old changelog entries are of any interest when you want to patch a system.

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -0,0 +1,16 @@
...
+In order to limit the amount of changelog metadata which is downloaded and to save disk space, you can put a limit on how many entries to keep on disk.
+
+This configuration option is in the [filename]``/etc/rhn/rhn.conf`` configuration file.
+The parameter defaults to [systemitem]``0``, which means unlimited.
+
+----
+java.max_changelog_entries = 0
+----
+
+When you set this parameter, it will come into effect only when packages are synchronized.
```
mcalmer: > When you set this parameter, it will come into effect only when packages are synchronized.

When you set this parameter, it will come into effect only for new packages synchronized.

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -0,0 +1,16 @@
...
+You might like to delete the cached data to remove older data.
```
mcalmer: I see L3s flying with the question "how can I remove cached data" :-)

```
$> spacewalk-sql -i
SQL> delete from rhnPackageRepodata;
SQL> insert into rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
(select sequence_nextval('rhn_repo_regen_queue_id_seq'),
        C.label,
        'package summary modification',
        'Y'
   from rhnChannel C);
SQL> \q
$>
```

WARNING: this will regenerate all metadata and this is a very long running task. Expect several hours until it is finished. It depends on the number of channels the customer have. The task run via taskomatic in background.

Loquacity:
Please review: https://github.com/SUSE/doc-susemanager/pull/572

PR #558, opened 2019-07-05
WIP: Add Maintenance Window Content

Loquacity:
https://github.com/SUSE/spacewalk/issues/8235

### DIFF
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -8,8 +8,10 @@ This section contains a checklist for your maintenance window, with links to fur
 
 . Stop services.
 . Check if the configuration is still correct, and apply the highstate.
-. Apply the latest updates
+. Apply the latest updates.
+See xref:upgrade:server-upgrade.adoc[].
 . Update to the latest service pack, if required.
+See xref:upgrade:sp-migration.adoc[].
 . Reboot the server.
 . Check if the configuration is still correct, and apply the highstate.
 . Start any stopped services.
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -1,0 +1,8 @@
+[[maintenance-window]]
+= Maintenance Window Checklist
+
+If you work on a system with scheduled maintenance windows, you might find it difficult to remember all the things that you need to during that critical downtime.
+
+This section contains a checklist for your maintenance window, with links to further information on performing each of the steps, if you need it.
+
+Coming soon ...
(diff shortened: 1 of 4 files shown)


keichwa:
General remark first: We should clearly differentiate between server and clients (proxy is a special client and should think about that one, too).

I'll now add my remarks in a way that they will make sense for the server.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -0,0 +1,17 @@
+[[maintenance-window]]
+= Maintenance Window Checklist
+
+If you work on a system with scheduled maintenance windows, you might find it difficult to remember all the things that you need to during that critical downtime.
```
keichwa:
```suggestion
If you work on a {productname} Server with scheduled maintenance windows, you might find it difficult to remember all the things that you need to during that critical downtime.
```

modules/administration/pages/maintenance-window.adoc
```diff
@@ -0,0 +1,17 @@
...
+
+This section contains a checklist for your maintenance window, with links to further information on performing each of the steps, if you need it.
+
+
+. Stop services.
```
keichwa: Do we want to single out special services such as "spacewalk services"?  Of course, if you work on a system it is always good if you stop as many services as possible.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -0,0 +1,17 @@
...
+. Check if the configuration is still correct, and apply the highstate.
```
keichwa: applying highstate only makes sense if you maintain the suma server with Salt tool.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -0,0 +1,17 @@
...
+. Apply the latest updates.
+See xref:upgrade:server-upgrade.adoc[].
+. Update to the latest service pack, if required.
+See xref:upgrade:sp-migration.adoc[].
+. Reboot the server.
```
keichwa: That's "safe", but not always necessary. You can even update (parts of) the kernel with live patching.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -0,0 +1,17 @@
...
+. Check if the configuration is still correct, and apply the highstate.
```
keichwa: about highstate, see above.

modules/administration/pages/maintenance-window.adoc
Loquacity: The original had "SAP and databases", but when I thought about it, it's highly dependent on what you have running. Perhaps a general statement, though?

modules/administration/pages/maintenance-window.adoc
Loquacity: Agreed, but it's probably best to mention it.

Issue #567, opened 2019-07-10
On window width < 1024px the toggle menu (and the menu) is unreachable


ncounter:
https://www.suse.com/documentation/suse-manager-4/4.0/suse-manager/installation/install-intro.html

Width >= 1024px :+1: 
![Screenshot from 2019-07-10 08-47-56](https://user-images.githubusercontent.com/7080830/60947085-08807d80-a2f0-11e9-8e64-1a405bbdc84b.png)

Width < 1024px :-1: 
![Screenshot from 2019-07-10 08-48-13](https://user-images.githubusercontent.com/7080830/60947094-0cac9b00-a2f0-11e9-815b-924688774106.png)

PR #568, opened 2019-07-10
Fix header bar content: do not overlaps element on small screen

ncounter:
Fix: https://github.com/SUSE/doc-susemanager/issues/567


jcayouette:
Do not merge until @ncounter is done testing.

PR #571, opened 2019-07-12
Supported browsers

Loquacity:
Add back missing admonition

### DIFF
--- a/modules/installation/pages/general-requirements.adoc
+++ b/modules/installation/pages/general-requirements.adoc
@@ -64,5 +64,5 @@ In order to use the {webui} to manage your {susemgr} environment, you will need 
 {susemgr} is supported on:
 
 * Latest Firefox browser shipped with {slsa}
+* Latest Chrome browser on all operating systems
 * Latest Edge browser shipped with Windows
-* Latest Chrome/Chromium browser on all operating systems
--- a/modules/installation/pages/general-requirements.adoc
+++ b/modules/installation/pages/general-requirements.adoc
@@ -2,10 +2,6 @@
 = General Requirements
 
 
-
-
-
-
 Before you begin your installation, check that your environment meets these requirements:
 
 * Current {scc} organization credentials
@@ -57,3 +53,9 @@ endif::[]
 Download {slsa}{nbsp}15{nbsp}SP1 or higher from https://www.suse.com/products/server/download/[SUSE Linux Enterprise Server - Media Download]
 
 //TODO: What about Uyuni?
+
+[[installation-general-supportedbrowsers]]
+== Supported Browsers for the {susemgr} {webui}
+
+In order to use the {webui} to manage your {susemgr} environment, you will need to ensure you are running an up to date web browser.
+{susemgr} is supported on the latest {sle} and Unix-based web browsers, as well as Windows and MacOS.
--- a/modules/installation/pages/general-requirements.adoc
+++ b/modules/installation/pages/general-requirements.adoc
@@ -54,8 +54,15 @@ Download {slsa}{nbsp}15{nbsp}SP1 or higher from https://www.suse.com/products/se
 
 //TODO: What about Uyuni?
 
+
+
 [[installation-general-supportedbrowsers]]
 == Supported Browsers for the {susemgr} {webui}
 
 In order to use the {webui} to manage your {susemgr} environment, you will need to ensure you are running an up to date web browser.
-{susemgr} is supported on the latest {sle} and Unix-based web browsers, as well as Windows and MacOS.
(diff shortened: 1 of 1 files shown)


mcalmer:
I think we make this statement as we do not know if the firefox browsers we ship on all sles versions is also the latest available. It could be an older version.

But I think it would be no problem to say latest firefox on every os and latest shipped with sles.

But please check back with @ncounter @lneves12 and @paususe what they think about it.

modules/installation/pages/general-requirements.adoc
```diff
@@ -57,3 +53,16 @@ endif::[]
...
+[[installation-general-supportedbrowsers]]
+== Supported Browsers for the {susemgr} {webui}
+
+In order to use the {webui} to manage your {susemgr} environment, you will need to ensure you are running an up to date web browser.
+
+{susemgr} is supported on:
+
+* Latest Firefox browser shipped with {slsa}
+* Latest Edge browser shipped with Windows
+* Latest Chrome/Chromium browser on all operating systems
```
ncounter: It should really work without problems, but I actually don't remember if it is *officially* in the list of supported browsers.
AFAIK (as I wrote it [here](https://wiki.microfocus.com/index.php?title=SUSE_Manager/FAQ#Which_browser_and_version_are_supported_for_SUSE_Manager.3F)) only `latest SLE browsers (Firefox and Chrome) and latest Windows (Edge)` are.

@mcalmer , @paususe who can tell us if we want/should include Chromium or not?

mcalmer: AFAIK this is basically our decission if we can/want to support it or not. In the end you and Luis might want to decide if chromium creates problems or not.

paususe: > /FAQ#Which_browser_and_version_are_supported_for_SUSE_Manager.3F)) only `latest SLE browsers (Firefox and Chrome).

AFAIK, SLE does not provide Chrome or Chromium, it's only available from the "Server:Unstable" repository in OBS:
https://software.opensuse.org/package/chromium

But I'd like to know, do we actually test SUSE Manager with all the latest Firefox versions on all supported SLE (which is exactly what the sentence "latest SLE browsers (Firefox)" implies? Same for Windows: do we actually test SUSE Manager with a fully-updated Windows?

If the answer to those is "no", and we don't know any specific issue with Chrome or Chromium, we we can very well just say something like "you need a modern browser  such as Firefox or Microsoft Edge. Internet Explorer is not supported due to technical limitations". That does not exclude Chrome but it does not explicitly include it either.

ncounter: @paususe I am in the process of evaluating a [browserstack account](https://www.browserstack.com) and finally we will be able to test SUSE Manager against different browsers, but I am not yet there.
At the moment, no we are not testing Firefox at all. It's just a matter of *knowing* the engine behind the browser version, and using it daily. So far the most stable has always been Firefox, and I personally develop on top of it. Other developers use Chrome so those two should be fine. And Edge uses the same engine of Chrome IIRC. Last but not least we use the headless-Chrom browser in the testsuite environment, which easily leads us to support Chrome from my POV.

Back to @mcalmer statement, if I can decide what I'd like to support, I would mention only Firefox and Chrome. For marketing reasons I see the need of Edge: if we can omit it even better, if we have to keep it in the loop that's okayish.

paususe: Edge is not using the same engine of Chrome but its own (Trident). Edge Chromium (the new forthcoming version, no stable release yet) will use Blink too.

Loquacity: Perhaps the original statement is better then? Check the first commit and LMK.

ncounter: > Edge is not using the same engine of Chrome but its own (Trident). Edge Chromium (the new forthcoming version, no stable release yet) will use Blink too.

Ah, you're right!! Anyway it seems Edge works fine for us so far.

> Edge is a need on Windows, as some organizations (mine until 2 months ago, for instance) do not install any other browser than the ones provided with the OS.

Yeah, indeed. If it would be possible I would drop Edge, but apparently *it is not possible*, so I am fine to keep it in the list.

**Bottom line**: Firefox and Chrome are safe, Edge is okeyish and we want/need to keep it in the loop. Let's exclude anything else for now.

Maybe with a proper testing we will figure out we can support many more, but not for now, it would be like jumping into a black hole, and it's nothing we want at the moment.
@Loquacity please then remove anything else (Chromium, MacOS/Safari, etc) and we are fine.

@lneves12 anything against ^^^?

PR #572, opened 2019-07-15
Edit Tuning info about changelog metadata storage. 

Loquacity:
Edits per feedback from @mcalmer on https://github.com/SUSE/doc-susemanager/pull/551

### DIFF
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -18,7 +18,7 @@ You can delete cached data from the command line, using the SQL interface:
 
 [IMPORTANT]
 ====
-Deleting cached data can take a long time.
+This process can take a long time.
 Depending on the number of channels you have and the amount of data to be deleted, it can potentially take several hours.
 The task is run in the background by Taskomatic, so you can continue to use {productname} while the operation completes, however you should expect some performance loss.
 ====
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -2,7 +2,7 @@
 = Tuning Changelogs
 
 Some packages have a long list of changelog entries.
-This data is downloaded by default, but is very rarely useful information.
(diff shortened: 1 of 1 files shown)


modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,5 +12,25 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
...
 ----
 
-When you set this parameter, it will come into effect only when packages are synchronized.
+When you set this parameter, it will come into effect only for new packages as they are synchronized.
 You might like to delete the cached data to remove older data.
+You can delete cached data from the command line, using the SQL interface:
+
+[IMPORTANT]
+====
+Deleting cached data can take a long time.
```
mcalmer: Regeneration take long. Delete is fast.

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,5 +12,25 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
...
+
+[IMPORTANT]
+====
+This process can take a long time.
+Depending on the number of channels you have and the amount of data to be deleted, it can potentially take several hours.
+The task is run in the background by Taskomatic, so you can continue to use {productname} while the operation completes, however you should expect some performance loss.
+====
+
+----
+spacewalk-sql -i
```
moio: @mcalmer do we want to suggest use of direct SQL in our manuals?

mcalmer: It is the only way to force the re-generation which is needed.

The only other way would be if we write a script which does it and you can call it.
Would be an option.

Loquacity: I'm personally not against directly using SQL, as long as it's the simplest method to achieve the goal. If an easier method becomes available, we can update :)

keichwa: thanks for all your reviews!

PR #579, opened 2019-07-22
Content Review: Client Configuration Guide

Loquacity:
Pull Request 1/2


Loquacity:
I don't have a preference, I've worked with both standards at different times (which is probably why we now have a mix, sorry!). If the docs team want title case, then I'm happy to go with that. We just need to make a decision.

keichwa:
> We just need to make a decision.

In this case, I vote for title style.

Loquacity:
This is about the halfway mark, so I'm going to open this one for review, and start a new PR for the next half.

modules/client-configuration/pages/disconnected-setup.adoc
```diff
@@ -1,226 +1,147 @@
...
+RMT replaces the subscription management tool (SMT), which can be used on older {sle} installations.
 
+In a disconnected {productname} setup, RMT or SMT uses an external network to connect to {scc}.
+All software channels and repositories are synchronized to a removable storage device.
+The storage device can then be used to update the disconnected {productname} installation.
 
-.Offline Usage Scenario
-[NOTE]
-====
-SMT and RMT are not made for server cascades.
```
Loquacity: I can't really work out what this sentence is saying, but after some discussion with a local dev, perhaps it's trying to say that you can't use it if you have a proxy set up? @mcalmer are you able to help me here?

keichwa:
On 2019-08-05 10:06, Lana Brindley wrote:

> -[NOTE]
> -====
> -SMT and RMT are not made for server cascades.
> 
> I can't really work out what this sentence is saying, but after some
> discussion with a local dev, perhaps it's trying to say that you can't
> use it if you have a proxy set up? @mcalmer are you able to help me
> here?

It probably means that you cannot connect a RMT server to a RMT server.
A RMT server must connect to SCC.

-- 
Karl Eichwalder

modules/client-configuration/pages/autoyast.adoc
```diff
@@ -1,20 +1,18 @@
 [[client-cfg-autoinstallation-autoyast]]
-= {ay}
+= AutoYaSt
 
+When you install a {sle} client, there are a number of questions you need to answer.
+In order to automate installation, you can create an {ay} file with all the answers to those questions, so that no user intervention is required.
```
keichwa:
```suggestion
To automate installation, you can create an {ay} file with all the answers to those questions, so that no user intervention is required.
```

modules/client-configuration/pages/salt-client-proxy.adoc
```diff
@@ -18,26 +18,26 @@ By default, the username is [systemitem]``root``.
...
 . Click btn:[Bootstrap] to begin registration.
 When the bootstrap process has completed, your client will be listed at menu:Systems[System List].
 
 
 
 // TODO: Might need an 'unsupported' note? LKB 2019-05-01
 
-Instead of the {webui}, you may use the command line to register a client through a proxy.
-This procedure requires that you have installed the Salt package on the client before registration, and have the Advanced systems module activated.
+Instead of the {webui}, you can use the command line to register a minion through a proxy.
```
keichwa: Better avoid the term "minion", and use "Salt client" instead.

modules/client-configuration/pages/kickstart.adoc
```diff
@@ -1,16 +1,18 @@
 [[client-cfg-reg-with-bootstrap-kickstart]]
 = Kickstart
 
+When you install a {rhel} client, there are a number of questions you need to answer.
+In order to automate installation, you can create a {kickstart} file with all the answers to those questions, so that no user intervention is required.
```
keichwa:
```suggestion
To automate installation, you can create a {kickstart} file with all the answers to those questions, so that no user intervention is required.
```

Loquacity:
> On 2019-08-05 10:06, Lana Brindley wrote: -[NOTE] -==== -SMT and RMT are not made for server cascades. I can't really work out what this sentence is saying, but after some discussion with a local dev, perhaps it's trying to say that you can't use it if you have a proxy set up? @mcalmer are you able to help me here?
> It probably means that you cannot connect a RMT server to a RMT server. A RMT server must connect to SCC.

Oh, that makes sense, but it also seems kind of self-evident. Do we need to call this out?

modules/client-configuration/pages/salt-client-proxy.adoc
Loquacity: missed those. Thanks!

keichwa:

> Oh, that makes sense, but it also seems kind of self-evident. Do we need
> to call this out?

I fear it's in because users sometimes have those ideas—maybe, @mcalmer
can tell us more ;)

-- 
                           90409 Nürnberg, Germany

mcalmer:
@Loquacity I think this was to answer the questions why SUSE Manager cannot connect directly to SMT/RMT like to SCC. The reason is, that they do not provide all required APIs for doing it and this was by design (not made for cascades).
So disconnected mode via filesystem is really the only was to do it.

Just to prevent customer questions or be able to point to this doc.

Loquacity:
> @Loquacity I think this was to answer the questions why SUSE Manager cannot connect directly to SMT/RMT like to SCC. The reason is, that they do not provide all required APIs for doing it and this was by design (not made for cascades).
> So disconnected mode via filesystem is really the only was to do it.
> 
> Just to prevent customer questions or be able to point to this doc.

OK, thanks for the clarification. I added a note, what do you think? Is it clearer than what was there before?

PR #594, opened 2019-07-29
Monitoring Updates

Loquacity:
* Added service discovery info (from @cavalheiro )
* Renamed file to Monitoring 

Still to come:
* More details on the steps required (waiting on @juliogonzalez )

https://github.com/SUSE/spacewalk/issues/8330
https://bugzilla.suse.com/show_bug.cgi?id=1137345

### DIFF
--- a/modules/administration/pages/monitoring.adoc
+++ b/modules/administration/pages/monitoring.adoc
@@ -5,14 +5,17 @@ You can monitor your {productname} environment using Prometheus and Grafana.
 {productname} Server and Proxy are able to provide self-health metrics, or install and manage a limited number of Prometheus exporters on Salt clients.
 
 Prometheus is a monitoring tool that is used to record real-time metrics in a time-series database.
-Prometheus collects metrics using HTTP pulls, allowing for higher performance and scalability.
-Prometheus is an open-source software project, written in Go.
-Its source code is available at https://github.com/prometheus/.
+It is an open-source software project, written in Go.
+Metrics are collected using HTTP pulls, allowing for higher performance and scalability.
+For more information about Prometheus, see https://prometheus.io/docs/.
 
 Grafana is a tool for data visualization, monitoring, and analysis.
 It is used to create dashboards with panels representing specific metrics over a set period of time.
 Grafana is commonly used together with Prometheus, but also supports other data sources such as ElasticSearch, MySQL, PostgreSQL, and Influx DB.
 For more information about Grafana, see https://grafana.com/docs/.
+
+You need to install Prometheus and Grafana on a machine separate from the {productname} Server.
+We recommend you use a managed client as your monitoring server.
 
 Prometheus and Grafana packages are included in the SUSE Manager Client Tools for {sle}{nbsp}12 and {sle}{nbsp}15.
 
@@ -114,14 +117,14 @@ See https://github.com/boynux/squid-exporter.
 
 == Install and Configure Prometheus
 
-Prometheus is installed from a package, and needs configuration before you can use it to gather metrics.
+Prometheus is installed on your monitoring server from a package, and needs configuration before you can use it to gather metrics.
 
 
 
 === Installing Prometheus
 
 .Procedure: Installing Prometheus
-. Install the [package]``golang-github-prometheus-prometheus`` package:
+. On the monitoring server, install the [package]``golang-github-prometheus-prometheus`` package:
 +
 ----
 zypper in golang-github-prometheus-prometheus
@@ -176,7 +179,10 @@ See https://github.com/wrouesnel/postgres_exporter.
 
 Installing and configuring exporters is done using a Salt formula.
 
-When you have the exporters installed and configured, you can manually configure monitoring, or set up service discovery so that Prometheus automatically starts scraping metrics from systems as they are enabled.
+When you have the exporters installed and configured, you can enable and configure monitoring.
+You can do this directly through the {productname} {webui}, or set up service discovery.
+Service discovery instructs Prometheus to automatically scrape metrics from systems as they are enabled.
+
 
 == Enable and Configure Monitoring
 
@@ -207,7 +213,6 @@ image::monitoring_enable_services.png[scaledwidth=80%]
 image::monitoring_configure_formula.png[scaledwidth=80%]
 
 
-
 .Procedure: Enable Service Discovery
 
 [IMPORTANT]
@@ -216,7 +221,7 @@ This feature is a technical preview available in {productname} 4.0.2 and later.
 It should not be used in production systems.
 ====
 
-. Open the Prometheus static configuration file at [path]``/etc/prometheus/prometheus.yml``.
+. On the monitoring server, open the Prometheus static configuration file at [path]``/etc/prometheus/prometheus.yml``.
 . Add or update the scrape configurations section:
 +
 ----
--- a/modules/administration/pages/monitoring.adoc
+++ b/modules/administration/pages/monitoring.adoc
@@ -226,7 +226,7 @@ This feature is a technical preview available in {productname} 4.0.2 and later.
 It should not be used in production systems.
 ====
 
-. On the monitoring server, open the Prometheus static configuration file at [path]``/etc/prometheus/prometheus.yml``.
+. On the monitoring server, open the Prometheus static configuration file [path]``/etc/prometheus/prometheus.yml``.
 . Add or update the scrape configurations section:
 +
 ----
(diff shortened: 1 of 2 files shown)


paususe:
> Still to come:
>     * More details on the steps required (waiting on @juliogonzalez )

Maybe @cavalheiro knows better?

juliogonzalez:
Yes, as mentioned at bugzilla, monitoring developers should review first, and when they are happy, QAM and I can have a final validation.

mcalmer:
@cavalheiro and @mateiw are on vacation now. @renner is the only option now when he is back

modules/administration/pages/monitoring.adoc
```diff
@@ -1,26 +1,20 @@
...
+You can monitor your {productname} environment using Prometheus and Grafana.
+{productname} Server and Proxy are able to provide self-health metrics, or install and manage a limited number of Prometheus exporters on Salt clients.
 
-// I do not see why we should mention where this was started; we do not do this elsewhere
 Prometheus is a monitoring tool that is used to record real-time metrics in a time-series database.
-Unlike other monitoring systems, Prometheus collects metrics using HTTP pulls, allowing for higher performance and scalability.
-Prometheus is an open-source software project, mostly written in Go.
+Prometheus collects metrics using HTTP pulls, allowing for higher performance and scalability.
+Prometheus is an open-source software project, written in Go.
 Its source code is available at https://github.com/prometheus/.
```
renner: The actual source code of Prometheus itself is here: https://github.com/prometheus/prometheus

Maybe instead of linking the source code, a link to the documentation could be a bit more helpful? Github does have a diagram though and is linking the website at https://prometheus.io/.

modules/administration/pages/monitoring.adoc
```diff
@@ -1,26 +1,20 @@
...
-{productname} Server and Proxy are now able to provide self-health metrics, or install and manage a limited number of Prometheus exporters on managed client systems.
+You can monitor your {productname} environment using Prometheus and Grafana.
+{productname} Server and Proxy are able to provide self-health metrics, or install and manage a limited number of Prometheus exporters on Salt clients.
 
-// I do not see why we should mention where this was started; we do not do this elsewhere
 Prometheus is a monitoring tool that is used to record real-time metrics in a time-series database.
-Unlike other monitoring systems, Prometheus collects metrics using HTTP pulls, allowing for higher performance and scalability.
-Prometheus is an open-source software project, mostly written in Go.
+Prometheus collects metrics using HTTP pulls, allowing for higher performance and scalability.
+Prometheus is an open-source software project, written in Go.
```
renner: Here we have three sentences all starting with Prometheus, but maybe it's not an issue.

modules/administration/pages/monitoring.adoc
```diff
@@ -214,6 +208,31 @@ image::monitoring_configure_formula.png[scaledwidth=80%]
...
 
+.Procedure: Enable Service Discovery
+
+[IMPORTANT]
+====
+This feature is a technical preview available in {productname} 4.0.2 and later.
+It should not be used in production systems.
+====
+
+. Open the Prometheus static configuration file at [path]``/etc/prometheus/prometheus.yml``.
```
renner: Maybe we should make it really clear here that this needs to be done on the machine where Prometheus is installed and running?

modules/administration/pages/monitoring.adoc
```diff
@@ -182,7 +176,7 @@ See https://github.com/wrouesnel/postgres_exporter.
 
 Installing and configuring exporters is done using a Salt formula.
 
-
+When you have the exporters installed and configured, you can manually configure monitoring, or set up service discovery so that Prometheus automatically starts scraping metrics from systems as they are enabled.
```
renner: Here it does not really get clear how someone would `manually configure monitoring` exactly, I guess. Maybe it could be replaced with `configure Prometheus to scrape the exposed metrics`, even though this gives only a little bit more information?

modules/administration/pages/monitoring.adoc
Loquacity: Linking to the docs is better. It also mirrors the Grafana section.

modules/administration/pages/monitoring.adoc
Loquacity: passive voice to the rescue! :tada: :wink:

modules/administration/pages/monitoring.adoc
Loquacity: The idea is that there are several procedures below that statement: for manually configuring, and for service discovery. I'll update the wording to make that clearer.

modules/administration/pages/monitoring.adoc
```diff
@@ -120,14 +117,14 @@ See https://github.com/boynux/squid-exporter.
...
 
 
 === Installing Prometheus
 
 .Procedure: Installing Prometheus
-. Install the [package]``golang-github-prometheus-prometheus`` package:
+. On the monitoring server, install the [package]``golang-github-prometheus-prometheus`` package:
 +
 ----
 zypper in golang-github-prometheus-prometheus
```
renner: While this is the correct manual way of installing the package when logged in to the monitoring server, we could mention here that the package could also be installed using SUSE Manager itself in case the monitoring server is a managed client. There is several ways to install a package via SUSE Manager, for example schedule the installation from the UI or API, or create a package state to make sure the package is always installed and possibly updated with every application of the *highstate*.

The same would apply for the installation of Grafana as well.

modules/administration/pages/monitoring.adoc
```diff
@@ -182,6 +179,9 @@ See https://github.com/wrouesnel/postgres_exporter.
 
 Installing and configuring exporters is done using a Salt formula.
 
+When you have the exporters installed and configured, you can enable and configure monitoring.
+You can do this directly through the {productname} {webui}, or set up service discovery.
```
renner: Here from my perspective it is not yet clear enough what exactly we mean with "you can enable and configure monitoring", I would guess it rather means "the monitoring server (Prometheus) can start scraping or gathering the metrics from monitored systems".

Further we can surely install and configure exporters directly through the web UI, but configuring Prometheus to scrape metrics from managed clients requires config changes on the monitoring server, so it would require at least a configuration channel assignment and a new revision of the config file for each monitored system that should be added. What I mean is: it does not really work as *directly* as it sounds here.

modules/administration/pages/monitoring.adoc
```diff
@@ -1,26 +1,23 @@
...
+For more information about Prometheus, see https://prometheus.io/docs/.
 
 Grafana is a tool for data visualization, monitoring, and analysis.
 It is used to create dashboards with panels representing specific metrics over a set period of time.
 Grafana is commonly used together with Prometheus, but also supports other data sources such as ElasticSearch, MySQL, PostgreSQL, and Influx DB.
 For more information about Grafana, see https://grafana.com/docs/.
 
-A Prometheus package is included in the SUSE Manager Client Tools for {sle}{nbsp}12 and {sle}{nbsp}15.
+You need to install Prometheus and Grafana on a machine separate from the {productname} Server.
+We recommend you use a managed client as your monitoring server.
```
0rnela: Here, the explicit instruction that Salt Client is an example of manager client van be added, given that it (Salt Client) appears further down in the text.

modules/administration/pages/monitoring.adoc
```diff
@@ -120,14 +117,14 @@ See https://github.com/boynux/squid-exporter.
 
 == Install and Configure Prometheus
 
-Prometheus is installed from a package, and needs configuration before you can use it to gather metrics.
+Prometheus is installed on your monitoring server from a package, and needs configuration before you can use it to gather metrics.
```
0rnela: Here also, the explicit instruction that Salt Client is an example of manager client van be added, given that it (Salt Client) appears further down in the text.

modules/administration/pages/monitoring.adoc
Loquacity: Yeah, I think that's the general problem with this whole section, actually. We make it sound very straight forward, when it's in reality a fair bit more complicated. I updated the wording on that bit, anyway.

modules/administration/pages/monitoring.adoc
```diff
@@ -213,6 +218,30 @@ image::monitoring_enable_services.png[scaledwidth=80%]
...
 
+.Procedure: Enable Service Discovery
+
+[IMPORTANT]
+====
+This feature is a technical preview available in {productname} 4.0.2 and later.
+It should not be used in production systems.
+====
+
+. On the monitoring server, open the Prometheus static configuration file at [path]``/etc/prometheus/prometheus.yml``.
```
keichwa:
```suggestion
. On the monitoring server, open the Prometheus static configuration file [path]``/etc/prometheus/prometheus.yml``.
```

PR #596, opened 2019-07-29
salt clients now, no longer minions

keichwa:


modules/administration/pages/public-cloud.adoc
```diff
@@ -322,7 +322,7 @@ Then [command]``service osad stop``.
 The bootstrap should now run with a new system ID, rather than a duplicate.
 
 
-If you are onboarding Salt minion clones, then you will also need to check if they have the same Salt minion ID.
+If you are onboarding Salt client clones, then you will also need to check if they have the same Salt client ID.
```
Loquacity: I think this will still be a minion ID. Might be worth checking.

modules/installation/pages/proxy-setup.adoc
```diff
@@ -117,8 +117,8 @@ Otherwise it can be left empty.
 CA Password::
 Enter the password that was used for the certificate of your {productname} server.
 
-Do You Want to Use an Existing SSH Key for Proxying SSH-Push Salt Minions?::
-Use this option if you want to reuse a SSH key that was used for SSH-Push Salt minions on the server.
+Do You Want to Use an Existing SSH Key for Proxying SSH-Push Salt Clients?::
```
Loquacity: The interface might say minion.

keichwa:

> -If you are onboarding Salt minion clones, then you will also need to
> check if they have the same Salt minion ID.
> +If you are onboarding Salt client clones, then you will also need to
> check if they have the same Salt client ID.
>
> I think this will still be a minion ID. Might be worth checking.

Thanks for spotting it.  Will talk to the devs.  In the long or not to
long run, they must change user-visible texts, too ;^)

-- 
                           90409 Nürnberg, Germany

Loquacity:
> Thanks for spotting it. Will talk to the devs. In the long or not to long run, they must change user-visible texts, too ;^)

Yes, this was discussed at retro. There seems to be general consensus that Salt client is the right way to go, although there are some places in the code that might not be easy to change (I seem to recall MinionID being one of them). Also, Pau wanted to chase up with consultants to get their feedback, too.

modules/administration/pages/public-cloud.adoc
keichwa: ready to push

modules/installation/pages/proxy-setup.adoc
keichwa: ditto

keichwa:
> Yes, this was discussed at retro. There seems to be general consensus that Salt client is the right way to go, although there are some places in the code that might not be easy to change (I seem to recall MinionID being one of them). Also, Pau wanted to chase up with consultants to get their feedback, too.

Perfect :) Great people.  No need that I disturb the discussion now ;)

PR #599, opened 2019-07-29
Please remove warning notes for prerelease versions

paususe:
They do not add anything useful anymore and can only mislead the reader into thinking the current version of SUSE Manager / Uyuni is not ready for production.

This kind of warnings should only be in the very version where they are relevant, and be removed in the next version.

### DIFF
--- a/modules/ROOT/pages/release-notes-version-4.0.adoc
+++ b/modules/ROOT/pages/release-notes-version-4.0.adoc
@@ -209,6 +209,7 @@ We have added a prototype of real time search using Algolia.
 We are still developing an alternative for offline reading.
 
 
+The search feature of this beta release is a prototype and points to development and testing sources.
 [[new-ui]]
 === Antora Static Site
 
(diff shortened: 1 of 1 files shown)


Loquacity:
These are only in the Beta sections. You have to scroll a long way down before you find them.

paususe:
You have to scroll down but it's so prominent (orange /red box, marks, etc) that is immediately seen.

Please remove.

modules/ROOT/pages/release-notes-version-4.0.adoc
```diff
@@ -218,14 +209,6 @@ We have added a prototype of real time search using Algolia.
 We are still developing an alternative for offline reading.
```
keichwa:
```suggestion

The search feature of this beta release is currently a prototype and points to development and testing sources.
```

keichwa:
> These are only in the Beta sections. You have to scroll a long way down before you find them. By removing them, we are removing the history of our documentation, which could lead to confusion later on. If you _really_ want them less prominent, then keep the content, but remove the admonition formatting.

Because these admonition are sailing under a beta version tag, they can easily go away, IMO.

PR #622, opened 2019-08-06
Public Cloud Edits

Loquacity:
https://github.com/SUSE/spacewalk/issues/8951

More to come ...


keichwa:
@Loquacity I think this is now also in here: https://github.com/SUSE/doc-susemanager/pull/657.  I think we can close this PR.

PR #624, opened 2019-08-06
Move Salt timeouts to own page

moio:
First PR in a string to reorganize high-scalability content.

Intent here is to group timeouts under one umbrella, in particular because I plan to completely remove the scale-minions.adoc page they are in now:
 - this PR moves out the "Presence Ping Timeout" and "Salt SSH Clients" sections
 - "Salt Client Onboarding Rate" will go under "Operation Recommendations" in the upcoming "Big Scale Deployment" Salt page


moio:
Part of https://github.com/SUSE/spacewalk/issues/5217

moio:
Thanks, merged. Not sure whether we want this on 3.2 - I guess it makes more sense to keep all this restructuring for `master` only, right?

Loquacity:
> Thanks, merged. Not sure whether we want this on 3.2 - I guess it makes more sense to keep all this restructuring for `master` only, right?

Yeah, I think so. None of it really meets the threshold for backporting.

PR #630, opened 2019-08-08
Content Review - Client Cfg - Part 2

Loquacity:
https://github.com/SUSE/spacewalk/issues/8362


Loquacity:
> I do not know whether we can already provide info about rhel 8.

Not last I checked. I think we're a while away from having any kind of RHEL 8 updates in SUMA.

cbbayburt:
> I do not know whether we can already provide info about rhel 8.

You can track the status here if you'd like: https://github.com/SUSE/spacewalk/issues/6420

keichwa:
On 2019-08-09 11:25, Can Bulut Bayburt wrote:
>> I do not know whether we can already provide info about rhel 8.
> 
> You can track the status here if you'd like:
> https://github.com/SUSE/spacewalk/issues/6420

Thanks for the pointer!  It looks as if there is still some way to
go.

-- 
Karl Eichwalder

Loquacity:
And ... we're done!

PR #634, opened 2019-08-09
Large Scale Deployment - Line Edit

Loquacity:
Following up on https://github.com/SUSE/doc-susemanager/pull/626


modules/salt/pages/large-scale-tuning.adoc
```diff
@@ -0,0 +1,524 @@
...
+
+[cols="1,1"]
+|===
+| Description          | The amount of memory reserved for PostgreSQL shared buffers, which contain caches of database tables and index data.
+| Tune when            | RAM changes
+| Value default        | 25% of total RAM
+| Value recommendation | 25-40% of total RAM
+| Location             | [path]``/var/lib/pgsql/data/postgresql.conf``
+| After changing       | Check the Memory Usage section in this document.
+| Notes                | The value is formatted as `<N>MB`, where `<N>` is the number of MiB to allocate.
```
keichwa:
```suggestion
| Notes                | The value is formatted as `<N>MiB`, where `<N>` is the number of MiB to allocate.
```
General remark: It looks as if we should switch to MiB and GiB where MB and GB isn't meant.

modules/salt/pages/large-scale-tuning.adoc
```diff
@@ -0,0 +1,524 @@
...
+
+=== `work_mem`
+
+[cols="1,1"]
+|===
+| Description          | The amount of memory allocated by PostgreSQL every time a connection needs to do a sort or hash operation.
+                         Every connection (as specified by `max_connections`) might make use of an amount of memory equal to a multiple of `work_mem`.
+| Tune when            | Individual query operations are too slow, and value is below 5 MB
+| Value default        |
+| Value recommendation | 2-20 MB
```
keichwa:
```suggestion
| Value recommendation | 2-20 MiB
```

modules/salt/pages/large-scale-tuning.adoc
```diff
@@ -0,0 +1,524 @@
...
+
+=== `swappiness`
+
+[cols="1,1"]
+|===
+| Description          | How aggressively the kernel moves unused data from memory to the swap partition.
+                         Setting a lower parameter typically reduces swap usage and results in better performance, especially when RAM memory is abundant.
+| Tune when            | RAM increases, or swap is used when RAM memory is sufficient.
+| Value default        | 60
+| Value recommendation | 1-60. For 128 GB of RAM, 10 is expected to give good results.
```
keichwa:
```suggestion
| Value recommendation | 1-60. For 128 GiB of RAM, 10 is expected to give good results.
```

modules/salt/pages/large-scale.adoc
```diff
@@ -0,0 +1,177 @@
...
+
+== Hardware and Infrastructure
+
+Not all problems can be solved with better hardware, but choosing the right hardware is an absolute necessity for large scale deployments.
+
+
+The minimum requirements for the {productname} Server are:
+
+* Eight or more recent {x86_64} CPU cores.
+* 32{nbsp}GB RAM.
```
keichwa:
```suggestion
* 32{nbsp}GiB RAM.
```

modules/salt/pages/large-scale.adoc
```diff
@@ -0,0 +1,177 @@
...
+If the {productname} Server is virtualized, enable the `elevator=noop` kernel command line option, for optimal I/O performance.
+You can check the current status with [command]``cat /sys/block/<DEVICE>/queue/scheduler``.
+This command will display a list of available schedulers with the currently active one in brackets.
+To change the scheduler before a reboot, use [command]``echo noop > /sys/block/<DEVICE>/queue/scheduler``.
+
+The minimum requirements for the {productname} Proxy are:
+
+* One {productname} Proxy per 500-1000 clients, depending on available network bandwidth.
+* Two or more recent {x86_64} CPU cores.
+* 16{nbsp}GB RAM, and sufficient storage for caching.
```
keichwa:
```suggestion
* 16{nbsp}GiB RAM, and sufficient storage for caching.
```

modules/salt/pages/large-scale.adoc
```diff
@@ -0,0 +1,177 @@
...
+You can use this formula to calculate the download time:
+
+----
+Size of updates * Number of clients / Theoretical download speed / 60
+----
+
+For example, the total time needed to transfer 400{nbsp}MB of upgrades through a physical link speed of 1{nbsp}GB/s to 3000 clients:
+
+----
+400 MB  * 3000 / 119 MB/s / 60 = 169 min
```
keichwa: If we'd go for MiB, re-calculation might be neccessary.

modules/salt/pages/large-scale-tuning.adoc
moio: In this case I am talking about the formatting of the configuration option - ie. the exact characters that have to be in `postgresql.conf` - so `MiB` will just not work here.

modules/salt/pages/large-scale.adoc
```diff
@@ -0,0 +1,177 @@
...
+
+
+
+== Operation Recommendations
+
+This section contains a range of recommendations for large scale deployments.
+
+[NOTE]
+====
+Always start small and scale up gradually.
```
moio: @Loquacity I took the liberty to move this under the Operation Recommendations, as it's not really about hardware/infrastructure requirements.

Feel free to just revert commit f9241ca2  in case you want it back.

modules/salt/pages/large-scale-tuning.adoc
Loquacity: I'm always a bit torn on using MiB and GiB. They're technically more correct, but also less frequently used and therefore a source of confusion (I've literally had people raise bugs because they thought the i was a typo). For that reason, I tend to use MB and GB by default, and only use MiB and GiB where it would be incorrect not to.

modules/salt/pages/large-scale-tuning.adoc
moio: Here the problem is all Postgres configuration variables are really set in MB, so I would not use MiB to save the reader a conversion.

moio:
I added fixes for @admd's comments, especially hyperlinks everywhere.

I think this is finally good to merge now!

keichwa:
On 2019-08-09 10:13, Silvio Moioli wrote:

> Here the problem is all Postgres configuration variables are really
> set in MB, so I would not use MiB to save the reader a conversion.

Sounds reasonably.

-- 
Karl Eichwalder

PR #375, opened 2019-03-21
Adding package endpoint salt pillar content to 3.2

Loquacity:
"backport" of https://github.com/SUSE/doc-susemanager/pull/366

PR #366, opened 2019-03-15
RPM endpoint salt pillars

Loquacity:
https://github.com/SUSE/doc-susemanager/issues/331
https://github.com/SUSE/spacewalk/issues/6890

PR #656, opened 2019-08-14
Use correct pillar name

moio:
For some reason currently we use the incorrect pillar names `rpm_download_point_XXX` instead of `pkg_download_point_XXX`.

See the original PR for 4.0: https://github.com/SUSE/doc-susemanager/pull/366/files#diff-8bf787463eb899039a59878b8b2ce800R179
For 3.2: https://github.com/SUSE/doc-susemanager/pull/375/files#diff-ba2328059d5e5a16c5dcbdff39701c1fR473

This is just a find/replace, maybe we should also check if any other edits were lost in the process.

### DIFF
--- a/modules/salt/pages/salt-states.adoc
+++ b/modules/salt/pages/salt-states.adoc
@@ -169,41 +169,41 @@ If you do not adjust these pillars directly, {productname} will fall back to the
 ----
 base:
   '*':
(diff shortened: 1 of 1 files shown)


Loquacity:
Hrm, must have changed at some point.

PR #657, opened 2019-08-15
Public cloud #8951 ke

keichwa:
Azure documentation.

Now we need better alignment with the general public cloud document.

### DIFF
--- a/modules/administration/pages/public-cloud-azure.adoc
+++ b/modules/administration/pages/public-cloud-azure.adoc
@@ -176,5 +176,37 @@ service postgresql status
 
 
 
-=== SUSE Manager Configuration
+=== SUSE Manager Setup Configuration
 
+// REMARK Do we want to list the details here?  Or is such a general xref good enough?
+
+For the SUSE Manager setup configuration in general, see xref:installation:server-setup.adoc[], _procedure "{productname} Setup"_.
+
+Pay special attention to the following settings:
+
+. In the first dialog, select [guimenu]``Set up SUSE Manager from scratch``
+. In the next dialog, enter a valid mail address for the administrator
+. It is very important to remember the password given for SSL.
+Without this password no SUSE Manager Proxy Server can be installed and no other changes can be made to the certificate.
+For example, this certificate is eg used on all registered systems to communicate with SUSE Manager Server.
+. In the [guimenu]``Database Settings`` dialog, it is enough to provide a password.
+Make sure also to remember this password.
+
+With these settings the installation can be started.
+The installation will finish without further input.
+
+When the installation is finished, an administration user will be created automatically.
+For Azure the system administrator user is called [literal]``admin`` and the password is built as follows:
+
+. The first part can be found with:
++
+
+----
+azuremetadata --instance-name
+----
+. The second part is [literal]``-suma``
+
+Alternatively, you can check the [path]``/var/log/susemanager_firstuser.log`` file.
+
+This completes the installation of SUSE Manager.
+From now on the server behaves identical as a non-Azure SUSE Manager Server.
--- a/modules/administration/pages/public-cloud-azure.adoc
+++ b/modules/administration/pages/public-cloud-azure.adoc
@@ -1,0 +1,3 @@
+[[public-cloud-azure]]
+= SUSE Manager Server in Azure
+
--- a/modules/administration/pages/public-cloud-azure.adoc
+++ b/modules/administration/pages/public-cloud-azure.adoc
@@ -71,3 +71,20 @@ This should be restricted for other server in this private network.
 
 ==== Network
 
+. DHCP must not set the host name.
+Check [path]``/etc/sysconfig/network/dhcp`` if `DHCLIENT_SET_HOSTNAME` is set to [literal]``no``.
+Change it if needed:
++
+
+----
+DHCLIENT_SET_HOSTNAME="no"
+----
+. Create the needed DNS Zones in Azure first.
+For more information, see the documentation from Microsoft Azure.
+
+. On the command line, add the system to the [path]``/etc/hosts`` file; replace [literal]``<ip_address>`` with the IP address of the server:
++
+
+// hostname -i?
+----
+echo "<ip_address> $(hostname -f) $(hostname)" >> /etc/hosts
--- a/modules/administration/pages/public-cloud-azure.adoc
+++ b/modules/administration/pages/public-cloud-azure.adoc
@@ -1,111 +1,54 @@
 [[public-cloud-azure]]
-= SUSE Manager Server in Azure
+= {productname} with Azure
+
+You can use {productname} Server and Proxy with the Microsoft Azure public cloud.
+This section discusses what you will need for running {productname} in Azure, and how to set up your installation.
 
 
+== Configure the Azure Cloud Instance
 
-== Resource Requirements
-
-
-
-=== Image and VM (Virtual Machine)
-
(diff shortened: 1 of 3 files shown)


modules/administration/pages/public-cloud-azure.adoc
```diff
@@ -0,0 +1,212 @@
+[[public-cloud-azure]]
+= SUSE Manager Server in Azure
+
+
+
+== Resource Requirements
+
+
+
+=== Image and VM (Virtual Machine)
```
Loquacity: Ooh, I'm not a fan of the nested headings. Can we either delete some, or add some text in between?

Loquacity:
Hrm,  modules/administration/pages/public-cloud-azure.adoc needs some editing. Do you want me to take the branch and update now, or should I wait until you have more feedback? There seem to be a lot of todos in the text still.

keichwa:
On 2019-08-16 02:56, Lana Brindley wrote:
> Hrm,  modules/administration/pages/public-cloud-azure.adoc needs some
> editing. Do you want me to take the branch and update now, or should I
> wait until you have more feedback? There seem to be a lot of todos in
> the text still.

Yes, please take the branch an edit.  Thanks for your help!

I hope MB will have time to review and help, too.

-- 
Karl Eichwalder

modules/administration/pages/public-cloud-azure.adoc
```diff
@@ -1,111 +1,54 @@
 [[public-cloud-azure]]
-= SUSE Manager Server in Azure
+= {productname} with Azure
```
keichwa: @Loquacity  @jcayouette Can we expand macros in chapter titles (now?)?

I'm also not sure whether all this will work with Uyuni.  Probably in general, but I did not check all the details.

Loquacity: I thought that was an old restriction. Better get @jcayouette 's advice on that.

keichwa:

> -= SUSE Manager Server in Azure
> += {productname} with Azure
>
> I thought that was an old restriction. Better get @jcayouette 's
> advice on that.

@loquacity You are probably right.  World is changing too fast ;)

-- 
                           90409 Nürnberg, Germany

keichwa:
@mbrookhuis I hope you are happy with our achievement --thanks for your help!  We'll ask you to do the technical review later :)

modules/administration/pages/public-cloud-azure.adoc
```diff
@@ -0,0 +1,164 @@
...
+When you create your Azure virtual machine, choose the `B2ms` size.
+The `B2ms` VM contains:
+
+* 2{nbsp}vCPUs
+* 8{nbsp}GB RAM
+* Up to 4 data disks
+* Max IOPS 2400
+* Temporary storage disk of 16{nbsp}GB.
+Data on this disk will be destroyed after the guest has been switched off.
+
```
jgleissner: This isn't really consistent with the general hardware requirements listed in the Installation Guide, which lists 16 GB for base and 32 GB for production systems as well as a minimum of 4 cores.

modules/administration/pages/public-cloud-azure.adoc
```diff
@@ -0,0 +1,164 @@
...
+mount /var/spacewalk
+chown -R wwwrun:root /var/spacewalk
+mount /var/lib/pgsql
+chown -R postgres:postgres /var/lib/pgsql
+mv /var/cache /var/cache.old
+mkdir /var/cache
+mount /var/cache
+rm -r /var/cache.old
+----
+
```
jgleissner: Any particular reason why the user shouldn't use the `suma-storage` script as described in the main public cloud page?

modules/administration/pages/public-cloud-azure.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+== Configure {productname} Server
+
+Ensure that your {productname} Server is registered with {scc}.
+
+When your server is registered, install these extra packages:
+----
+zypper -n in spacecmd mlocate sysstat
+----
```
jgleissner: Please note that `spacecmd` has been added to the SUSE Manager Server images (though not released yet). Any particular reason why `mlocate` and `sysstat` are required?

modules/administration/pages/public-cloud-azure.adoc
mbrookhuis: This is disk /dev/sdb. This is a temporary disk on Azure. I think we should not use this. Data is volatile. This is coming from Azure and is present on almost every system. For me this is a confusing disk on Azure. If you select you VMSize you will see a Temporary Storage column.

mbrookhuis: And this is my test system. For customers I would go for at least 4 CPUs and at least 16 Gb of RAM. Sorry for not being that clear

modules/administration/pages/public-cloud-azure.adoc
mbrookhuis: We had some cases where we had performance problems. A deeper research showed that the root cause was having /var/spacewalk and /var/lib/pgsql on the same disk (could be different volumes). Also take in consideration that we normally but the backup of the database in the /var/spacewalk directory under backup-db. Postgress uses this also for storing the journals. So every operation against the database is written in the database and in the journal. When they are both on the same disk this could cause a bottleneck. But I have never tested/researched this with SSD disks. 
Another reason is that using both on the same disk increases the change of running out of diskspace eg when downloading new channels/products. And you don't want your database running out of diskspace....

modules/administration/pages/public-cloud-azure.adoc
mbrookhuis: That is correct. The document was written on the moment it wasn't. I forgot that also spacewalk-utils is needed (eg for taskotop which is a useful utility to see what taskomatic is doing). 

mlocate is just a tool I often use the search for files. And yes you can also use "find", but like locate and most customer I worked with share that feeling.

Sysstat helps you troubleshooting performance issues or to check the general health. Also this is often requested by support. Still fighting with PM to get this on every SLES as default (for SLES4SAP it is).

mbrookhuis:
I am currently working with a real customer to install SUSE Manager in Azure. If you wish I could share the documentation I am writing for that customer. This will contain a complete installation including staging, registering and later also salt.

mbrookhuis:
Another note. In my doc there is also a Public address configured. This will not be done in real production environments.

modules/administration/pages/public-cloud-azure.adoc
jgleissner: Safeguarding the data base from running out of disk space sounds definitely like a good idea. I guess suma-storage should be changed to create multiple partitions for that purpose on the storage volume, and possibly have support for multiple storage volumes added, in case somebody wants to set it up that way for performance reasons.

Regarding the potential performance issue, it's certainly valuable information, but I think it may require more research before we make specific recommendations in the documentation. As a side note, according to https://docs.microsoft.com/en-us/azure/virtual-machines/windows/disks-types, a small 64GB premium SSD volume doesn't really give you a better performance over a standard SSD volume of that size (maybe apart from a slightly lower latency). It'll cost more though.

keichwa:
Michael Brookhuis <notifications@github.com> writes:

> Another note. In my doc there is also a Public address configured. This
> will not be done in real production environments.

I removed it; see https://github.com/SUSE/doc-susemanager/pull/new/public-clould-feedback-ke

-- 
                           90409 Nürnberg, Germany

PR #666, opened 2019-08-21
Restructure Public Cloud specific information

jgleissner:
Since the public cloud specific information is not actually about administration but setting up the system, I moved the existing information into the Installation guide and also integrated them into the appropriate chapters, so admins in public clouds can simply go through the Installation guide without jumping back and forth in the docs.


Loquacity:
I would like to hold off on doing this kind of major architecture change until after the content review is complete, please. Additionally, I'm not convinced this belongs in the Install Guide, I will need to look closer at it, and I can't do that until after 4.0.1 submission happens. Sorry to be the bad guy here :(

modules/installation/pages/public-cloud-registration.adoc
```diff
@@ -0,0 +1,21 @@
+[[public-cloud-registration]]
+= Registration in the Public Clouds
+
+Since {productname} is preinstalled in Public Cloud images, there is no installation step required. However, {productname} Server needs to be registered with {scc} in order to receive updates.
```
keichwa:
```suggestion
Because {productname} is preinstalled in Public Cloud images, there is no installation step required.
However, {productname} Server needs to be registered with {scc} to receive updates.
```

Loquacity:
Closing this PR because master has moved on a bit. Will create a new PR.

Loquacity:
https://github.com/SUSE/doc-susemanager/pull/724

PR #681, opened 2019-08-26
Add Canonical IP safeguard

paususe:
For IP reasons, add note about Canonical not supporting or endorsing SUSE Manager. We should have had that since the very beginning:
https://ubuntu.com/legal/intellectual-property-policy

Also, make a bit more straightforward the first part of the page.

### DIFF
--- a/modules/client-configuration/pages/clients-ubuntu.adoc
+++ b/modules/client-configuration/pages/clients-ubuntu.adoc
@@ -4,19 +4,11 @@
 
 This section contains information about registering Salt clients running {ubuntu} operating systems.
 
-Support for {ubuntu} Clients was added in {susemgr} 3.2.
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/clients-ubuntu.adoc
```diff
@@ -4,19 +4,11 @@
 
 This section contains information about registering Salt clients running {ubuntu} operating systems.
 
-Support for {ubuntu} Clients was added in {susemgr} 3.2.
+{susemgr} supports {ubuntu} 16.04 LTS and 18.04 LTS Clients using Salt. Traditional clients are not supported.
```
Loquacity:
```suggestion
{susemgr} supports {ubuntu} 16.04 LTS and 18.04 LTS Clients using Salt. 
Traditional clients are not supported.
```

PR #687, opened 2019-08-27
update info on supported clients

keichwa:
https://github.com/SUSE/spacewalk/issues/9189
https://bugzilla.suse.com/show_bug.cgi?id=1148175

### DIFF
--- a/modules/installation/pages/client-requirements.adoc
+++ b/modules/installation/pages/client-requirements.adoc
@@ -23,7 +23,7 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
 | {sle} 11 | {x86}, {x86}_64, {ipf}, IBM {ppc}, {ibmz}  | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 7 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 6 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
-| {sls} for SAP 15 | {x86}_64, {ppc}                    | icon:check[role="green"]      | icon:check[role="green"]
+| {sls} for SAP | {x86}_64, {ppc}                       | icon:check[role="green"]      | icon:check[role="green"]
 | {rhel} 8 | {x86}_64                                   | icon:question[role="gray"]    | icon:question[role="gray"]
 | {rhel} 7 | {x86}_64                                   | icon:check[role="green"]      | icon:check[role="green"]
 | {rhel} 6 | {x86}, {x86}_64                            | icon:check[role="green"]      | icon:check[role="green"]
--- a/modules/installation/pages/client-requirements.adoc
+++ b/modules/installation/pages/client-requirements.adoc
@@ -23,13 +23,15 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
 | {sle} 11 | {x86}, {x86}_64, {ipf}, IBM {ppc}, {ibmz}  | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 7 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 6 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
-| {sls} for SAP | {x86}_64, {ppc}                       | icon:check[role="green"]      | icon:check[role="green"]
+| {sls} for SAP 15 | {x86}_64, {ppc}                    | icon:check[role="green"]      | icon:check[role="green"]
 | {rhel} 8 | {x86}_64                                   | icon:question[role="gray"]    | icon:question[role="gray"]
 | {rhel} 7 | {x86}_64                                   | icon:check[role="green"]      | icon:check[role="green"]
 | {rhel} 6 | {x86}, {x86}_64                            | icon:check[role="green"]      | icon:check[role="green"]
-| CentOS 7 | {x86}, {x86}_64                            | icon:question[role="gray"]    | icon:question[role="gray"]
-| CentOS 6 | {x86}, {x86}_64                            | icon:question[role="gray"]    | icon:question[role="gray"]
+| CentOS 7*) | {x86}, {x86}_64                          | icon:question[role="gray"]    | icon:question[role="gray"]
+| CentOS 6*) | {x86}, {x86}_64                          | icon:question[role="gray"]    | icon:question[role="gray"]
 | {opensuse} Leap 15.1 | {x86}_64                       | icon:times[role="danger"]      | icon:check[role="green"]
 | {ubuntu} 16.04 | {x86}_64                             | icon:times[role="danger"]      | icon:check[role="green"]
 | {ubuntu} 18.04 | {x86}_64                             | icon:times[role="danger"]      | icon:check[role="green"]
 |===
+
+*) If it is part of the Expanded Support offering.


modules/installation/pages/client-requirements.adoc
```diff
@@ -23,13 +23,15 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
 | {sle} 11 | {x86}, {x86}_64, {ipf}, IBM {ppc}, {ibmz}  | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 7 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
 | {sls}-ES 6 | {x86}_64                                 | icon:check[role="green"]      | icon:check[role="green"]
-| {sls} for SAP | {x86}_64, {ppc}                       | icon:check[role="green"]      | icon:check[role="green"]
+| {sls} for SAP 15 | {x86}_64, {ppc}                    | icon:check[role="green"]      | icon:check[role="green"]
```
mcalmer: 12 and 11 are supported as well. This was the reason why we removed the version.
I think PM need to come up with a rule about the granulartity of product names and versions.

modules/installation/pages/client-requirements.adoc
```diff
@@ -23,13 +23,15 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
...
 | {rhel} 8 | {x86}_64                                   | icon:question[role="gray"]    | icon:question[role="gray"]
 | {rhel} 7 | {x86}_64                                   | icon:check[role="green"]      | icon:check[role="green"]
 | {rhel} 6 | {x86}, {x86}_64                            | icon:check[role="green"]      | icon:check[role="green"]
-| CentOS 7 | {x86}, {x86}_64                            | icon:question[role="gray"]    | icon:question[role="gray"]
-| CentOS 6 | {x86}, {x86}_64                            | icon:question[role="gray"]    | icon:question[role="gray"]
+| CentOS 7*) | {x86}, {x86}_64                          | icon:question[role="gray"]    | icon:question[role="gray"]
+| CentOS 6*) | {x86}, {x86}_64                          | icon:question[role="gray"]    | icon:question[role="gray"]
```
mcalmer: If it is RES, he is supported (green). This line was meant for plain CentOS.

Loquacity:
I agree with MC. Making this more specific is just going to make it harder to maintain.

modules/installation/pages/client-requirements.adoc
keichwa: ok, I reverted that one.

modules/installation/pages/client-requirements.adoc
keichwa: @joesusecom we please need your feedback.

keichwa:
> I agree with MC. Making this more specific is just going to make it harder to maintain.

Yes.  And I'd even rather vote to maintain this list out of our books.  It belongs on some product page on suse.com that we'd reference from our books.  Within our books this list is always "wrong" because most of the time it is out of date.

mcalmer:
Maybe we should use a common approach for products and OSes.

If we put numbers for SLES, CentOS and others, we should do it also for other products and OSes.
AFAIK we tried to remove the versions everywhere, but it was not accepted by customers and PM.

So we added them again to some products and now we get one bugreport after the other about this.

modules/installation/pages/client-requirements.adoc
mcalmer: Oh, i just see another problem. Expanded Support for version 7 is only available for x86_x64.
If we mean here just plain CentOS 7 the arch list might be correct. Together with Expanded Support it is wrong.

paususe:
Let's think of it from the customer point of view. I have:
* SLES server for SAP and some other workloads
* Probably some server running RHEL or RES because some application is only certified for that
* A handful of lab or toy servers running openSUSE or CentOS because they are not critical
* Desktops running Ubuntu
* Desktops running Windows
* A handful of Macs for the marketing and graphical design guys

For Windows, you have InTune or SCCM, so not something we should worry about.

For Mac, well, we don't support it and I don't know if we should in the future.

Now it comes to the Linux side: there is this product called SUSE Manager which says it supports SLES, RHEL (and they even have something called RES that they support!), Ubuntu and more. Wooohooo! This rules! Let's buy it!

But when I try to manage my systems, it's all trouble:
* Everything works on SLES
* RHEL support works for 6 and 7 but for 8 there is no modular support (yet)
* RES is only for x86-64
* CentOS has no CVE audit support
* Ubuntu currently supports half the features but nobody told me
...

And that's exactly why we need a feature matrix like I asked in https://github.com/SUSE/spacewalk/issues/8304

It may be part of the "Client Book", or it may be part of some other book, but we definitely need to go for OS + version + feature is supported yes/no. Otherwise, Customers, Sales, Consulting and of course PM will complain (and I will perfectly understand them).

Loquacity:
> > I agree with MC. Making this more specific is just going to make it harder to maintain.
> 
> Yes. And I'd even vote to maintain this list out of our books. It belongs on some product page on suse.com that we'd reference from our books. Within our books this list is always "wrong" because most of the time it is out of date.

I've proposed this in the past and got voted off the island.

Loquacity:
> Maybe we should use a common approach for products and OSes.
> 
> If we put numbers for SLES, CentOS and others, we should do it also for other products and OSes.
> AFAIK we tried to remove the versions everywhere, but it was not accepted by customers and PM.
> 
> So we added them again to some products and now we get one bugreport after the other about this.

I suspect we will get endless bug reports about this forever. I'm kind of just resigned to it now.

Loquacity:
> And that's exactly why we need a feature matrix like I asked in [SUSE/spacewalk#8304](https://github.com/SUSE/spacewalk/issues/8304)
> 
> It may be part of the "Client Book", or it may be part of some other book, but we definitely need to go for OS + version + feature is supported yes/no. Otherwise, Customers, Sales, Consulting and of course PM will complain (and I will perfectly understand them).

I think this is a valid perspective, but I'm not optimistic that adding one more thing that may or may not be out of date is going to reduce the bug reports. We will add the matrix, I just  don't expect it to be a silver bullet.

paususe:
> > > I agree with MC. Making this more specific is just going to make it harder to maintain.
> > 
> > 
> > Yes. And I'd even vote to maintain this list out of our books. It belongs on some product page on suse.com that we'd reference from our books. Within our books this list is always "wrong" because most of the time it is out of date.
> 
> I've proposed this in the past and got voted off the island.

@keichwa @Loquacity 

What place in suse.com, outside the SUSE Manager books, would you propose? 

Who would maintain the support matrix?

How would it be published? 

Now that we have control over suse.com/documentation, it's hard for me to see a better place than our books.

Loquacity:
> > > > I agree with MC. Making this more specific is just going to make it harder to maintain.
> > > 
> > > 
> > > Yes. And I'd even vote to maintain this list out of our books. It belongs on some product page on suse.com that we'd reference from our books. Within our books this list is always "wrong" because most of the time it is out of date.
> > 
> > 
> > I've proposed this in the past and got voted off the island.
> 
> @keichwa @Loquacity
> 
> What place in suse.com, outside the SUSE Manager books, would you propose?

https://www.suse.com/lifecycle/

> 
> Who would maintain the support matrix?
> 
> How would it be published?

I think the matrix falls into our remit because it's about features, not products.

> 
> Now that we have control over suse.com/documentation, it's hard for me to see a better place than our books.

For most things, yes. For the support stuff I think it moves too quickly and is too detailed for documentation. Also, it's entirely outside of our control, and our domain of knowledge. We're always reacting: people raise bugs telling us "X is supported" or "Y is no longer supported" and we just go :woman_shrugging: and update accordingly. Seems inefficient to me.

keichwa:
I'll withdraw my PR.  This will be address as part of https://github.com/SUSE/spacewalk/issues/8304 .

 @Loquacity  I'm wondering whether I can simply close this PR here?

paususe:
> > @keichwa @Loquacity
> > What place in suse.com, outside the SUSE Manager books, would you propose?
> 
> https://www.suse.com/lifecycle/

I'm not sure that's the best place.

We are talking about what SUSE Manager supports, not about what SUSE supports. I. e. if feature X of SUSE Manager works on RHEL7, or feature Y of SUSE Manager works on SLES12.

> > Who would maintain the support matrix?
> > How would it be published?
> 
> I think the matrix falls into our remit because it's about features, not products.
> 
> > Now that we have control over suse.com/documentation, it's hard for me to see a better place than our books.
> 
> For most things, yes. For the support stuff I think it moves too quickly and is too detailed for documentation. Also, it's entirely outside of our control, and our domain of knowledge. We're always reacting: people raise bugs telling us "X is supported" or "Y is no longer supported" and we just go  and update accordingly. Seems inefficient to me.

But if we are going to maintain it anyway, what's the benefit of having it under suse.com/lifecycle instead of suse.com/documentation (that we already control and can update in a snap) ?

paususe:
> I'll withdraw my PR. This will be address as part of [SUSE/spacewalk#8304](https://github.com/SUSE/spacewalk/issues/8304) .
> 
> @Loquacity I'm wondering whether I can simply close this PR here?

@keichwa 
Until #8304 is done (that can take some time), I think we should merge your changes here. A bit of information is better than no information.

keichwa:
Ok, but then the only change is CentOS support. @mcalmer commented, and my current proposal is not correct.  It is unclear what @joesusecom want us to list here.  We depend on Joe's feedback.  Moving to block for the moment.

Loquacity:
> 
> But if we are going to maintain it anyway, what's the benefit of having it under suse.com/lifecycle instead of suse.com/documentation (that we already control and can update in a snap) ?

Yes, that's why I lost the argument ;)

I think you've recognised the problem here. Documentation isn't the right place, but nor is anywhere else.

Loquacity:
> Ok, but then the only change is CentOS support. @mcalmer commented, and my current proposal is not correct. It is unclear what @joesusecom wants us to list here. We depend on Joe's feedback. Moving it to blocked for the moment.

It might be best to close this one without merging for now, otherwise it'll get way out of date.

keichwa:
closing per feedback.
wondering whether https://github.com/SUSE/spacewalk/issues/9189 should stay open? From there I already referenced https://github.com/SUSE/spacewalk/issues/8304.  #9189 is linked with a bug that we probably should close, too ("worksforme").

Loquacity:
> closing per feedback.
> wondering whether [SUSE/spacewalk#9189](https://github.com/SUSE/spacewalk/issues/9189) should stay open? From there I already referenced [SUSE/spacewalk#8304](https://github.com/SUSE/spacewalk/issues/8304). #9189 is linked with a bug that we probably should close, too ("worksforme").

Yeah, go ahead and close #9189 and the bug. #8304 is gonna take some work ...

PR #688, opened 2019-08-28
Admin Guide: Security Hardening section

Loquacity:
https://github.com/SUSE/spacewalk/issues/7792

### DIFF
--- a/modules/administration/pages/security-hardening.adoc
+++ b/modules/administration/pages/security-hardening.adoc
@@ -1,0 +1,55 @@
+[[security-hardening]]
+= Security Hardening
+
+While no implementation is ever completely secure from bad actors, there are some things you can do within {productname} to improve your security.
+
+This section contains a number of suggestions for hardening your security posture.
+The primary focus is to reduce the likelihood that a bad actor could access Salt clients in the case of a breach of the {productname} Server.
+In this case, we assume that the Salt master is running on the {productname} Server.
+
+
+To begin, there are some built-in mechanisms that make the {productname} Server less vulnerable to bad actors.
+The Salt master runs as a non-privileged user, and the Salt API only listens on localhost, not an external address.
+
+Additionally, Salt clients cannot send jobs to the Salt master in the default configuration.
+This can be manually adjusted, but it is not recommended.
+
+There are a limited number of ports required for clients to communicate with the server, which simplifies network management.
+The only required ports are 4505 and 4506.
+Ports 80 and 443 are only used for package repositories, and they secured with SSL trusted certificates and GPG keys.
+
+The server never initiates a TCP handshake with a client.
+Clients initiate every connection, and the communication is bi-drectional after initiation.
+
+Each client creates a set of RSA keys.
+The public key is accepted by the server when a new system is added.
+If the keys are secure on the server, only clients with the correct private keys can be authenticated.
+The keys should be secured by specific users.
+
+For extra key security, you can implement a key policy.
+A key policy that allows only pre-seeded keys on the master is the most secure.
+Additionally, you can implement a key policy that allows clients to communicate only with a master that has a specific digital fingerprint.
+
+You can also implement a key size limit to increase security.
+
+All data between the server and clients is encrypted with a shared AES key.
+The key is is regenerated every 24 hours, or whenever a client is removed.
+
+Client pillar data is also encrypted with AES.
+In this case, a new AES session key is created every time a pillar is refreshed.
+
+Running Salt commands as root may be avoided by additional configuration in Salt. This allows for administrators to run salt commands as his or her user id and enables better auditing.
+
+If implementing non-root users, additional security may be added such as only allowing specific salt commands. Salt even allows for limiting the passing of specific predicates (arguments) to those limited commands.
+
+User authentication to Active Directory or some other authentication server such as some other LDAP server may be configured for command line commands by enabling external authentication.
+
+Of course {productname} may authenticate to an LDAP server with RBAC.
+
+Jobs issued through {productname} will be logged in the database. Jobs issues at the command line will be logged in /var/log/salt/master. Jobs may be logged to a remote server by implementing an external master job cache.
+
+We need to get engineering's approval, but showed at SaltConf integration the {productname} with SaltStack Enterprise. This may provide additional management, auditing, integration, and oversite.
+
+In some situations (multi-master, multi-tenant) Salt masters may be managed by an upstream master. This would need to be implemented carefully on a {productname} system, but could allow for remote management of security settings, reacting to any unauthorized changes, and enforcement of the proper state of the {productname}/Salt Master system. This should also get approval from engineering.
+
+We could dive deeper into each of these things if you think the customer needs it, but at a high level there are many proactive actions that may be taken to avoid a security breach.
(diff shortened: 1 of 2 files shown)


modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
+[[security-hardening]]
+= Security Hardening
+
+While no implementation is ever completely secure from bad actors, there are some things you can do within {productname} to improve your security.
+
+This section contains a number of suggestions for hardening your security posture.
+The primary focus is to reduce the likelihood that a bad actor could access Salt clients in the case of a breach of the {productname} Server.
+In this case, we assume that the Salt master is running on the {productname} Server.
+
+To begin, there are some built-in mechanisms that make the {productname} Server less vulnerable to bad actors.
```
keichwa: the next sentence is unrelated to this sentence (it's not a built-in mechanism and it would be the same on any other server system).  What war the built-in mechanisms?

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Additionally, Salt clients cannot send jobs to the Salt master in the default configuration.
+This can be manually adjusted, but it is not recommended.
+
+There are a limited number of ports required for clients to communicate with the server, which simplifies network management.
+The only required ports are 4505 and 4506.
+Ports 80 and 443 are only used for package repositories, and they secured with SSL trusted certificates and GPG keys.
+
+The server never initiates a TCP handshake with a client.
+Clients initiate every connection, and the communication is bi-drectional after initiation.
```
keichwa:
```suggestion
Clients initiate every connection, and the communication is bi-directional after initiation.
```

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+To begin, there are some built-in mechanisms that make the {productname} Server less vulnerable to bad actors.
+The Salt master runs as a non-privileged user, and the Salt API only listens on localhost, not an external address.
+
+Additionally, Salt clients cannot send jobs to the Salt master in the default configuration.
+This can be manually adjusted, but it is not recommended.
+
+There are a limited number of ports required for clients to communicate with the server, which simplifies network management.
+The only required ports are 4505 and 4506.
+Ports 80 and 443 are only used for package repositories, and they secured with SSL trusted certificates and GPG keys.
```
mcalmer: Port 80 is not secured by SSL. I havn't checked if you can disable it.

modules/administration/pages/security-hardening.adoc
mcalmer: Exception: salt-ssh or ssh-push . Here the server initiate a ssh connection to the client.

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+The server never initiates a TCP handshake with a client.
+Clients initiate every connection, and the communication is bi-drectional after initiation.
+
+Each client creates a set of RSA keys.
+The public key is accepted by the server when a new system is added.
+If the keys are secure on the server, only clients with the correct private keys can be authenticated.
+The keys should be secured by specific users.
```
mcalmer: All this is for "salt" clients, but the title is generic. So we should either say here explicit salt or the whole hardening guide should concentrate on salt.

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+For extra key security, you can implement a key policy.
+A key policy that allows only pre-seeded keys on the master is the most secure.
+Additionally, you can implement a key policy that allows clients to communicate only with a master that has a specific digital fingerprint.
```
mcalmer: I have no idea what this mean and I think customers will have the same question.

How to implement "a key policy"?
How to set it up/configure it?

If this is a guide which should explain how to make a server more secure, we need examples for everything!

As a sales guy say: You need to provide examples for everything. If you don't, customers become creative.

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+You can also implement a key size limit to increase security.
```
mcalmer: How?

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+All data between the server and clients is encrypted with a shared AES key.
+The key is is regenerated every 24 hours, or whenever a client is removed.
+
+Client pillar data is also encrypted with AES.
+In this case, a new AES session key is created every time a pillar is refreshed.
+
+
+You do not need to run Salt commands as root.
+This requires some additional configuration, so that administrators can run Salt commands using their user ID.
+Adjusting this also allows for improved auditing.
```
mcalmer: How to set this up? Examples!

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+You can add additional user-level controls, such as only allowing specific Salt commands.
+When you have selected a subset of allowed Salt commands, you can further limit the arguments that are accepted for each command.
```
mcalmer: How to configure this?

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+You can configure additional user authentication mechanisms, by connecting to an Active Directory or other user authentication mechanism.
+You can also enable external authentication methods like LDAP for command line access.
+When you set up LDAP, it is also possible to connect to the LDAP server using RBAC.
```
mcalmer: How, how and how?

modules/administration/pages/security-hardening.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Jobs issued through the {productname} {webui} are logged in a database.
+Jobs issues at the command line are logged to [path]``/var/log/salt/master``.
+Jobs can also be logged to a remote server, by using an external master job cache.
```
mcalmer: And finally : how ?

modules/administration/pages/security-hardening.adoc
Loquacity: That was me trying to organise the content better, failing, and then forgetting to remove the old sentence :confused:

Loquacity:
> Sorry, but I have always the same question:-)
> 
> We need to say "how" they can configure this. Otherwise we will get exactly this question back from the customer.
> 
> More details and **Examples** needed.

My thoughts exactly.

modules/administration/pages/security-hardening.adoc
Loquacity: I suspect this is all done with Salt states. Can someone on the Salt squad confirm?

modules/administration/pages/security-hardening.adoc
Loquacity: As per previous, I think this would be done with states, so the relevant xref is to https://opensource.suse.com/doc-susemanager/suse-manager/salt/salt-states.html

mcalmer:
I think we need to document which config variables a customer need to change, what OS configurations he need to do, etc.

@brejoc @dincamihai @meaksh @moio does somebody of you have deeper knowledge and can help with concrete examples?

moio:
Not really

Loquacity:
Any objections to merging this as is? I have no doubt that we will get feedback on it and can iterate.

mcalmer:
I am not happy with this, as it raise a lot of questions and answer none.
From my point of view we must provide configuration examples when we say in a guide that something should be configured. But it seems we have no idea how to do it. So better not release it until we have examples.

Loquacity:
> I am not happy with this, as it raise a lot of questions and answer none.
> From my point of view we must provide configuration examples when we say in a guide that something should be configured. But it seems we have no idea how to do it. So better not release it until we have examples.

OK, in that case it's probably best to close this PR. We can revisit later if we get more info.

PR #690, opened 2019-08-30
Upgrade postgres #9203 ke

keichwa:
https://github.com/SUSE/spacewalk/issues/9203
https://bugzilla.suse.com/show_bug.cgi?id=1148342

### DIFF
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -3,7 +3,7 @@
 
 This is about PostgreSQL database migration.
 
-If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based PostgreSQL version 10.
+If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based on PostgreSQL version 10.
 If you are already using PostreSQL version 10, skip this section.
 If you are using an older version, such as version 9.6, you must migrate PostgreSQL to version 10 before you begin the database migration.
 For more information about migrating PostgreSQL 9.6 to version 10, see xref:db-migration.adoc#db-migration-prepare[].
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -5,7 +5,7 @@ This is about PostgreSQL database migration.
 
 If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based PostgreSQL version 10.
 If you are already using PostreSQL version 10, skip this section.
-If it is based on a older one such as version 9.6 you must migrate PostgreSQL to version 10 first.
+If you are using an older version, such as version 9.6, you must migrate PostgreSQL to version 10 before you begin the database migration.
 For more information about migrating PostgreSQL 9.6 to version 10, see xref:db-migration.adoc#db-migration-prepare[].
 If you still run PostgreSQL 9.4 on {productname} 3.2 see the 3.2 product documentation at https://www.suse.com/documentation/suse-manager-3/3.2/susemanager-best-practices/html/book.suma.best.practices/bp.sp.migration.html#sp.migration.postgresql.
 Direct migration from PostgreSQL 9.4 to version 10 is not supported.
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -58,7 +58,7 @@ A regular upgrade will create a complete copy of the database, so you will need 
 Regular upgrades can take a considerable amount of time, depending on the size of the database and the speed of the storage system.
 Regular upgrades also require database downtime, your database will not be accessible while the upgrade is running.
 
-A *fast upgrade* only takes a few minutes, and uses no additional disk space.
+A fast upgrade only takes a few minutes, and uses no additional disk space.
 However, fast upgrades are more likely to fail, and if this occurs you must restore the database from the backup.
 
 
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -4,7 +4,7 @@
 This is about PostgreSQL database migration.
 
 If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based PostgreSQL version 10.
-If yes, you can skip this section.
+If you are already using PostreSQL version 10, skip this section.
 If it is based on a older one such as version 9.6 you must migrate PostgreSQL to version 10 first.
 For more information about migrating PostgreSQL 9.6 to version 10, see xref:db-migration.adoc#db-migration-prepare[].
 If you still run PostgreSQL 9.4 on {productname} 3.2 see the 3.2 product documentation at https://www.suse.com/documentation/suse-manager-3/3.2/susemanager-best-practices/html/book.suma.best.practices/bp.sp.migration.html#sp.migration.postgresql.
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -7,7 +7,7 @@ If you want to upgrade {productname} 3.2 to {productname} 4 you must check whe
 If you are already using PostreSQL version 10, skip this section.
 If you are using an older version, such as version 9.6, you must migrate PostgreSQL to version 10 before you begin the database migration.
 For more information about migrating PostgreSQL 9.6 to version 10, see xref:db-migration.adoc#db-migration-prepare[].
-If you still run PostgreSQL 9.4 on {productname} 3.2 see the 3.2 product documentation at https://www.suse.com/documentation/suse-manager-3/3.2/susemanager-best-practices/html/book.suma.best.practices/bp.sp.migration.html#sp.migration.postgresql.
+If you run PostgreSQL{nbsp}9.4 on {productname} 3.2, see the 3.2 product documentation: https://www.suse.com/documentation/suse-manager-3/3.2/susemanager-best-practices/html/book.suma.best.practices/bp.sp.migration.html#sp.migration.postgresql.
 Direct migration from PostgreSQL 9.4 to version 10 is not supported.
 
 
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -64,8 +64,6 @@ With the fast upgrade the risk to run out of disk space is smaller.
 
 In case of failure you do not have the implicit backup that you get when doing a regular upgrade.
 The regular upgrade will copy the database files instead of hardlinking them.
-This statement will scare customers and they will tend to unnecessarily waste time and space by avoiding the fast upgrade.
-
 
 .Procedure: Performing a Regular Upgrade
 . Perform a database backup.
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -1,6 +1,11 @@
 [[db-migration]]
 = Database Migration
 
+If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based PostgreSQL version 10.  If yes, you can skip this section.  If it is based on a older version such as 9.6 you must migrate PostgreSQL to version 10 first.
+
+
+
+
 
 When you installed {productname} Server 3.1 or earlier on {sls}{nbsp}12{nbsp}SP2 or earlier, it was based on a PostgreSQL{nbsp}9.4 database.
 PostgreSQL{nbsp}9.4 can be upgraded to PostgreSQL{nbsp}9.6.
--- a/modules/upgrade/pages/db-migration.adoc
+++ b/modules/upgrade/pages/db-migration.adoc
@@ -59,7 +59,12 @@ Regular upgrades can take a considerable amount of time, depending on the size o
 Regular upgrades also require database downtime, your database will not be accessible while the upgrade is running.
 
 A fast upgrade only takes a few minutes, and uses no additional disk space.
-However, fast upgrades are more likely to fail, and if this occurs you must restore the database from the backup.
+However, if a fast upgrade fails, you must restore the database from the backup.
+With the fast upgrade the risk to run out of disk space is smaller.
+
+In case of failure you do not have the implicit backup that you get when doing a regular upgrade.
+The regular upgrade will copy the database files instead of hardlinking them.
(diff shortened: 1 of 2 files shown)


modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
 [[db-migration]]
 = Database Migration
 
+This is about PostgreSQL database migration.
 
-When you installed {productname} Server 3.1 or earlier on {sls}{nbsp}12{nbsp}SP2 or earlier, it was based on a PostgreSQL{nbsp}9.4 database.
-PostgreSQL{nbsp}9.4 can be upgraded to PostgreSQL{nbsp}9.6.
+If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based PostgreSQL version 10.
```
Loquacity:
```suggestion
If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based on PostgreSQL version 10.
```

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
+If yes, you can skip this section.
```
Loquacity:
```suggestion
If you are already using PostreSQL version 10, skip this section.
```

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
+If it is based on a older one such as version 9.6 you must migrate PostgreSQL to version 10 first.
```
Loquacity:
```suggestion
If you are using an older version, such as version 9.6, you must migrate PostgreSQL to version 10 before you begin the database migration.
```

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
+For more information about migrating PostgreSQL 9.6 to version 10, see xref:db-migration.adoc#db-migration-prepare[].
+If you still run PostgreSQL 9.4 on {productname} 3.2 see the 3.2 product documentation at https://www.suse.com/documentation/suse-manager-3/3.2/susemanager-best-practices/html/book.suma.best.practices/bp.sp.migration.html#sp.migration.postgresql.
```
Loquacity:
```suggestion
If you run PostgreSQL{nbsp}9.4 on {productname} 3.2, see the 3.2 product documentation: https://www.suse.com/documentation/suse-manager-3/3.2/susemanager-best-practices/html/book.suma.best.practices/bp.sp.migration.html#sp.migration.postgresql.
```

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
+Direct migration from PostgreSQL 9.4 to version 10 is not supported.
```
Loquacity:
```suggestion
You cannot migrate directly from PostgreSQL{nbsp}9.4 to version 10.
```

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
 ====
 
+PostgreSQL upgrades can be performed in two ways: a regular upgrade, or a fast upgrade:
 
-
-== Upgrade PostgreSQL
-
-PostgreSQL upgrades can be performed in two ways: a regular upgrade, or a fast upgrade.
-A regular upgrade will create a complete copy of the database, so you will need double the existing database size of space available.
+A *regular upgrade* will create a complete copy of the database, so you will need double the existing database size of space available.
```
Loquacity: I'm not a fan of the emphasis here, can we write around it? Perhaps use a def list or something?

modules/upgrade/pages/migrate-3x-4x.adoc
```diff
@@ -24,19 +24,25 @@ We strongly recommend that you contact SUSE Consulting to assist with this proce
...
+====
+It is important that PostgreSQL 10 is already running on your {productname} 3.2 system.
+For more information, see xref:db-migration.adoc[].
+====
+
 During migration, the database on the source system needs to get exported, or dumped.
 The dump is compressed, and temporarily stored on the target system.
 The compression is done using [command]``gzip`` using the default compression options.
-Maximum compression only yields about 10% of space savings.
+Maximum compression only yields about 10 % of space savings.
```
Loquacity: Chicago (9.18, 16th ed) says no space between the numeral and the % symbol.

modules/upgrade/pages/migrate-3x-4x.adoc
```diff
@@ -24,19 +24,25 @@ We strongly recommend that you contact SUSE Consulting to assist with this proce
...
 Before you begin, use this command on the source system to check the size of the database:
 ----
 du -sch /var/lib/pgsql/data
 ----
 
-Ensure you have at least 30% of the total database size available in [path]``/var/spacewalk/tmp`` on the target system.
+Ensure you have at least 30 % of the total database size available in [path]``/var/spacewalk/tmp`` on the target system.
```
Loquacity: ibid.

modules/upgrade/pages/db-migration.adoc
mantel: I'm not a native English speaker but I always thought that "must" is considered impolite and one should use "you will need to" instead. But if it has Lana's blessing it will be ok :)

modules/upgrade/pages/db-migration.adoc
mantel: I wonder whether we should mention 9.4 at all in the first place. Quite some time ago we introduced some change that requires a "new" feature of postgresql that was not available in 9.4. So if somebody is still running this ancient postgresql version, he definitely also is running an totally outdated version of SUSE Manager. I think 9.4 is unusable for SUSE Manager since longer than a year.

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
 Regular upgrades can take a considerable amount of time, depending on the size of the database and the speed of the storage system.
 Regular upgrades also require database downtime, your database will not be accessible while the upgrade is running.
 
-A fast upgrade only takes a few minutes, and uses no additional disk space.
-However, fast upgrades are more likely to fail, and if this occurs you will need to restore the database from backup.
+A *fast upgrade* only takes a few minutes, and uses no additional disk space.
+However, fast upgrades are more likely to fail, and if this occurs you must restore the database from the backup.
```
mantel: No, this is utterly wrong and needs to get changed! The fast upgrade is NOT "more likely to fail", in fact it is even more safe as your risk to run out of disk space is smaller. However it is more dangerous in the sense that in case of failure you do not have the implicit backup you get when doing a regular upgrade (which will copy the database files instead of hardlinking them).
This statement will scare customers and they will tend to unnecessarily waste time and space by avoiding the fast upgrade.

modules/upgrade/pages/db-migration.adoc
```diff
@@ -84,161 +88,14 @@ The [path]``pg-migrate.sh`` script performs these operations:
 If the upgrade fails, the migration script will attempt to restore the database to its original state.
 ====
 
-
-
 .Procedure: Performing a Fast PostgreSQL Upgrade
 . Perform a database backup.
+Without a verified database backup, you must not initiate a fast upgrade.
```
mantel: Again I'm amazed by the "you must not" :) I would have worded it as "you should not" or even "it is strongly recommended..."

keichwa: Yes, sometimes passive voice is better.

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
 ====
 
+PostgreSQL upgrades can be performed in two ways: a regular upgrade, or a fast upgrade:
 
-
-== Upgrade PostgreSQL
-
-PostgreSQL upgrades can be performed in two ways: a regular upgrade, or a fast upgrade.
-A regular upgrade will create a complete copy of the database, so you will need double the existing database size of space available.
+A *regular upgrade* will create a complete copy of the database, so you will need double the existing database size of space available.
```
keichwa:
```suggestion
A regular upgrade will create a complete copy of the database, so you will need double the existing database size of space available.
```

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
 Regular upgrades can take a considerable amount of time, depending on the size of the database and the speed of the storage system.
 Regular upgrades also require database downtime, your database will not be accessible while the upgrade is running.
 
-A fast upgrade only takes a few minutes, and uses no additional disk space.
-However, fast upgrades are more likely to fail, and if this occurs you will need to restore the database from backup.
+A *fast upgrade* only takes a few minutes, and uses no additional disk space.
```
keichwa:
```suggestion
A fast upgrade only takes a few minutes, and uses no additional disk space.
```

modules/upgrade/pages/db-migration.adoc
keichwa: agreed. i remove it.

modules/upgrade/pages/migrate-3x-4x.adoc
keichwa: thanks for looking it up!

keichwa:
```suggestion
Maximum compression only yields about 10% of space savings.
```

modules/upgrade/pages/migrate-3x-4x.adoc
keichwa:
```suggestion
Ensure you have at least 30% of the total database size available in [path]``/var/spacewalk/tmp`` on the target system.
```

modules/upgrade/pages/db-migration.adoc
keichwa: @mantel This is technical documentation (not marketing speech), Hubert!
We probably should define somewhere what it actually means if we say "must" or "should".

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,65 @@
...
-== Upgrade PostgreSQL
-
-PostgreSQL upgrades can be performed in two ways: a regular upgrade, or a fast upgrade.
 A regular upgrade will create a complete copy of the database, so you will need double the existing database size of space available.
 Regular upgrades can take a considerable amount of time, depending on the size of the database and the speed of the storage system.
 Regular upgrades also require database downtime, your database will not be accessible while the upgrade is running.
 
 A fast upgrade only takes a few minutes, and uses no additional disk space.
-However, fast upgrades are more likely to fail, and if this occurs you will need to restore the database from backup.
+However, fast upgrades are more likely to fail, and if this occurs you must restore the database from the backup.
```
keichwa:
```suggestion
However, if a fast upgrade fails, you must restore the database from the backup.
With the fast upgrade the risk to run out of disk space is smaller.

In case of failure you do not have the implicit backup that you get when doing a regular upgrade.
The regular upgrade will copy the database files instead of hardlinking them.
This statement will scare customers and they will tend to unnecessarily waste time and space by avoiding the fast upgrade.
```

modules/upgrade/pages/db-migration.adoc
keichwa: thanks for clarification, Hubert.  I hope it is better worded now.

modules/upgrade/pages/db-migration.adoc
```diff
@@ -1,61 +1,70 @@
...
 Regular upgrades also require database downtime, your database will not be accessible while the upgrade is running.
 
 A fast upgrade only takes a few minutes, and uses no additional disk space.
-However, fast upgrades are more likely to fail, and if this occurs you will need to restore the database from backup.
+However, if a fast upgrade fails, you must restore the database from the backup.
+With the fast upgrade the risk to run out of disk space is smaller.
+
+In case of failure you do not have the implicit backup that you get when doing a regular upgrade.
+The regular upgrade will copy the database files instead of hardlinking them.
+This statement will scare customers and they will tend to unnecessarily waste time and space by avoiding the fast upgrade.
```
mantel: This ("will scare customers") was just a comment by me. It was not meant to go into the document :)

modules/upgrade/pages/db-migration.adoc
keichwa: I see but I nevertheless would like to mention this case.  Of course, if developers or supporter would think this will cause more confusion than help, I can remove it.

modules/upgrade/pages/db-migration.adoc
keichwa: thanks for noting.  I forgot to remove it ;)

modules/upgrade/pages/db-migration.adoc
mantel: Honestly I do not care. IMHO it is just a no-op by now. I bet nobody is using 9.4 any longer.

modules/upgrade/pages/db-migration.adoc
Loquacity: We use passive voice in technical writing a lot, because the actor is often the least important information. So in this case, it's more about word order than active/passive. Here, we need the reader to notice this sentence if they don't have a backup, so we put that information first. If they have a backup, they can skip this sentence. If they don't, we need them to keep reading: do not initiate the upgrade.

As for the word choice, "should" is not as definite as "must". I often sit on the couch and say "I should clean the floors", but it doesn't mean I do. The same can be said for "it is strongly recommended". Aside from being far too wordy, it's still not as strong as "must". You don't get to argue with "must" ;)

All that said, it could be even clearer by saying "do not". This is both more concise, and more definite. But do you notice how I just removed the actor entirely? The actor is simply not important in this sentence, which further backs up the case for passive voice.

I wrote guidance on this type of thing here: https://github.com/SUSE/doc-susemanager/wiki/wug#language

mantel: Thanks for the free English lesson :)

I was under the impression that - exactly because "must" is so strong - it should not be used in English as it can be perceived as impolite and one should use something like "you might consider to do this and that" instead.

The British are too polite to be honest,
while the Germans are too honest to be polite.

:)

Loquacity: > Thanks for the free English lesson :)

Always happy to (over)share the knowledge :stuck_out_tongue: 

> 
> I was under the impression that - exactly because "must" is so strong - it should not be used in English as it can be perceived as impolite and one should use something like "you might consider to do this and that" instead.

That would probably be true if we were in a different setting. 

> 
> The British are too polite to be honest,
> while the Germans are too honest to be polite.

:joy:

And technical writers are the least polite of all. None of those extra words hanging around and messing the place up for us!

PR #691, opened 2019-08-30
mgr-sync logging

keichwa:
Info about logging taken from the manpage.
@mcalmer does it make sense this way?  The previous version looked like a cut-and-paste error to me.

### DIFF
--- a/modules/reference/pages/command-line-tools.adoc
+++ b/modules/reference/pages/command-line-tools.adoc
@@ -702,8 +702,7 @@ Show help message and exit.
 With [command]``mgr-sync`` you may add or synchronize products and channels.
 The [command]``mgr-sync`` command also enables and refreshes SCC data.
 
-By default [command]``mgr-sync`` writes basic debug information to
-[path]``/var/log/rhn/mgr-sync.log``.
+By default, [command]``mgr-sync`` writes basic debug information to [path]``/var/log/rhn/mgr-sync.log``.
 Further debugging information can be obtained by either with [option]``--debug`` or adding [literal]``mgrsync.debug = <DEBUGLEVEL>`` to [path]``/etc/rhn/rhn.conf``.
 
 [IMPORTANT]
--- a/modules/reference/pages/command-line-tools.adoc
+++ b/modules/reference/pages/command-line-tools.adoc
@@ -703,7 +703,7 @@ With [command]``mgr-sync`` you may add or synchronize products and channels.
 The [command]``mgr-sync`` command also enables and refreshes SCC data.
 
 By default, [command]``mgr-sync`` writes basic debug information to [path]``/var/log/rhn/mgr-sync.log``.
-Further debugging information can be obtained by either with [option]``--debug`` or adding [literal]``mgrsync.debug = <DEBUGLEVEL>`` to [path]``/etc/rhn/rhn.conf``.
+Get more debugging information with [option]``--debug`` or by adding [literal]``mgrsync.debug = <DEBUGLEVEL>`` to [path]``/etc/rhn/rhn.conf``.
 
 [IMPORTANT]
 .Admin credentials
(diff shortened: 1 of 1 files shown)


modules/reference/pages/command-line-tools.adoc
```diff
@@ -702,12 +702,9 @@ Show help message and exit.
...
 
-
-By default, log rotation of [path]``spacewalk-repo-sync`` is disabled.
-This logfile is run by a cron job, and managed by a configuration file.
-You can enable log rotation by editing the configuration file at [path]``etc/sysconfig/rhn/reposync`` and setting the [command]``MAX_DAYS`` parameter.
-This will permit the deletion of files that are older than the specified period of time.
-
+By default [command]``mgr-sync`` writes basic debug information to
+[path]``/var/log/rhn/mgr-sync.log``.
+Further debugging information can be obtained by either with [option]``--debug`` or adding [literal]``mgrsync.debug = <DEBUGLEVEL>`` to [path]``/etc/rhn/rhn.conf``.
```
mcalmer: ~/.mgr-sync overwrite values from rhn.conf. Something initialize it with 
```
mgrsync.debug = ""
```
so a value in rhn.conf has no effect. Maybe we just document `.mgr-sync` in the users home directory?

modules/reference/pages/command-line-tools.adoc
```diff
@@ -702,12 +702,9 @@ Show help message and exit.
...
 The [command]``mgr-sync`` command also enables and refreshes SCC data.
 
-
-By default, log rotation of [path]``spacewalk-repo-sync`` is disabled.
-This logfile is run by a cron job, and managed by a configuration file.
-You can enable log rotation by editing the configuration file at [path]``etc/sysconfig/rhn/reposync`` and setting the [command]``MAX_DAYS`` parameter.
-This will permit the deletion of files that are older than the specified period of time.
-
+By default [command]``mgr-sync`` writes basic debug information to
+[path]``/var/log/rhn/mgr-sync.log``.
```
Loquacity: This doesn't need a linebreak.

modules/reference/pages/command-line-tools.adoc
```diff
@@ -702,12 +702,9 @@ Show help message and exit.
 With [command]``mgr-sync`` you may add or synchronize products and channels.
 The [command]``mgr-sync`` command also enables and refreshes SCC data.
 
-
-By default, log rotation of [path]``spacewalk-repo-sync`` is disabled.
-This logfile is run by a cron job, and managed by a configuration file.
-You can enable log rotation by editing the configuration file at [path]``etc/sysconfig/rhn/reposync`` and setting the [command]``MAX_DAYS`` parameter.
-This will permit the deletion of files that are older than the specified period of time.
-
+By default [command]``mgr-sync`` writes basic debug information to
```
Loquacity:
```suggestion
By default, [command]``mgr-sync`` writes basic debug information to
```

modules/reference/pages/command-line-tools.adoc
Loquacity:
```suggestion
Get more debugging information with [option]``--debug`` or by adding [literal]``mgrsync.debug = <DEBUGLEVEL>`` to [path]``/etc/rhn/rhn.conf``.
```

keichwa: I'm not sure what I should do (it at all).  ~/.mgr-sync is not mentioned in the mgr-sync manpage.  Later we can enhance the mgr-sync description here, but ATM I just want to get rid of obviously wrong statements ;)

mcalmer: Trust the code:-)
The problem is, that somebody read it , try it and will fail.

Maybe you can add at least one comment that ~/.mgr-sync exists and may overwrite values defined in rhn.conf. If something is not working, check that file.

keichwa: ok :)

keichwa:
.mgr-sync is now in )plud minog .  please review again.

PR #701, opened 2019-09-09
Custom certificates

Loquacity:
https://github.com/SUSE/spacewalk/issues/3698
https://bugzilla.suse.com/show_bug.cgi?id=1144552


modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
 
+You can use custom certificates with {productname} and {productname} Proxy.
 
+This section covers how to use a third party SSL certificate authority with a new {productname} installation, and replacing existing certificate with new custom certificates.
 
+Before you begin, ensure you have:
 
+* A certificate authority (CA) SSL public certificate
+* An SSL private key
+* An SSL public certificate
```
mcalmer: I would name it more explicit "Server Certificate" and "Server Key".

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
 
+Your key and certificate files must be in PEM format.
 
-The following section will guide you through using a custom certificate with {productname} {productnumber} and {susemgrproxy} {productnumber}.
+The host name of the SSL keys and certificates must match the host name of the machine you deploy them on.
```
mcalmer: New RFCs more or less requires that the host name is configured in the "X509v3 Subject Alternative Name" Section of the Certificate. This is very technical but one of the typical problems of self generated Certificates.

In addition we could say that in that section you can also list multiple hostnames if the server has them.

mcalmer: Like below, we should say that hostnames in the certificates must be FQDNs

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
 ----
-. Execute {productname} setup with
+Ensure that the [systemitem]``set-cname`` parameter is the fully-qualified domain name of your {productname} Server.
+You can use the the [systemitem]``set-cname`` parameter multiple times if you require multiple aliases.
+. Install the RPM that contains the newly generated certificate:
 +
-
 ----
-yast2 susemanagersetup
+rpm -Uhv /root/ssl-build/lnx0259a/rhn-org-httpd-ssl-key-pair-lnx0259a-1.0-2.noarch.rpm
```
mcalmer: The version/release number will be increased with every time you re-create the certificates.
Not sure if we need to tell that to our customers that they need to look for the latest version of the RPM?

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
 
+== Re-Create Existing Certificates
 
-After completing {yast}
-firstboot procedures, export your current environment variables and point them to the correct SSL files to be imported.
-Running these commands will make the default certificate obsolete after executing the [command]``yast2 susemanagersetup`` command.
-For more information on {yast}
- firstboot, see https://www.suse.com/documentation/suse-manager-3/singlehtml/suse_manager21/book_susemanager_install/book_susemanager_install.html#sec.manager.inst.setup.
-[[bp.cert.custom.setup.proc]]
+If your existing custom certificates have expired or stopped working for any reason, you can generate a new certificate from the existing CA.
```
mcalmer:
```suggestion
If your existing custom certificates have expired or stopped working for any reason, you can generate a new server certificate from the existing CA.
```

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
+configure-proxy.sh
+----
+. At the ``Do you want to import existing certificates?`` prompt, type kbd:[y]
+. Follow the prompts to complete setup.
 
 
-[[bp.cert.custom.setup]]
-== Setup
 
+== Re-Create Existing Certificates
```
mcalmer:
```suggestion
== Re-Create Existing Server Certificates
```

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
 
-After completing {yast}
-firstboot procedures, export your current environment variables and point them to the correct SSL files to be imported.
-Running these commands will make the default certificate obsolete after executing the [command]``yast2 susemanagersetup`` command.
-For more information on {yast}
- firstboot, see https://www.suse.com/documentation/suse-manager-3/singlehtml/suse_manager21/book_susemanager_install/book_susemanager_install.html#sec.manager.inst.setup.
-[[bp.cert.custom.setup.proc]]
+If your existing custom certificates have expired or stopped working for any reason, you can generate a new certificate from the existing CA.
 
+.Procedure: Re-Creating an Existing Certificate
```
mcalmer:
```suggestion
.Procedure: Re-Creating an Existing Server Certificate
```

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
-+
+----
+spacewalk-service restart
+----
 
 
 
-[[bp.cert.custom.proxy]]
-== Using a Custom Certificate with {susemgrproxy}
+== Create New Certificates
```
mcalmer:
```suggestion
== Create and Replace CA and Server Certificates
```

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
 
+If you need to create entirely new certificates for an existing installation, you will need to delete the old certificates before you begin.
```
mcalmer: We should say that this is dangerous as it could happen that you by accident cut the trust chain from the clients to the server. If this happens, the admin need to login to every client and deploy the new CA there.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
+--set-org-unit="ORGANIZATION UNIT" --set-email="name@example.com" \
+--set-hostname="susemanager.example.top" --set-cname="example.com"
+----
+Ensure that the [systemitem]``set-cname`` parameter is the fully-qualified domain name of your {productname} Server.
+You can use the the [systemitem]``set-cname`` parameter multiple times if you require multiple aliases.
+. Generate the CA certificate RPM:
++
+----
+rhn-ssl-tool --gen-ca --rpm-only --dir="/root/ssl-build" \
+--from-ca-cert=`path_to_certificate_file`
```
mcalmer: I think this is not needed here. The commands above should automatically generate the RPMs.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
+----
+. Generate the server certificate RPM:
++
+----
+rhn-ssl-tool --gen-server --rpm-only --dir="/root/ssl-build" \
+--from-server-key=`path_to_server_key_file` \
+--from-server-cert=`path_to_server certificate_file`
```
mcalmer: I think this is not needed here. The commands above should automatically generate the RPMs.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
+----
+You will need to generate a server certificate RPM for each proxy, using their host names and cnames.
```
mcalmer: This could be a "Note" for step 3. The command of Step 3 is the one the customer should use to create new server certificates for the proxies.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
+. Install the CA certificate RPM on all clients by updating the channel.
+
+.Procedure: Deploying Certificates on Salt Clients
+
+// Does the above procedure also work for Salt? LKB 2019-09-10
+
+. On the client, in the {productname} {webui}, navigate to menu:Salt[Remote Commands].
+. In the first field, enter [guimenu]``salt-call state.apply certs``.
+In the second field, after the ``@``, enter [guimenu]``*``.
+Click btn:[Find Targets].
```
mcalmer: This is correct when you want to just deploy the new CA, but maybe we should just try out if a `highstate` on salt clients would do the same. I checked and the "certs" state is part of the highstate.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,195 @@
...
+. Check that the appropriate clients are displayed, and click btn:[Run Command] to deploy the certificates on the clients.
+
+image::deploy_cert_salt.png[scaledwidth=80%]
```
mcalmer: Hmm, there is the half missing.

- After the old and the new CA is trusted by every client the customer need to replace the server certificate on Server and Proxies.
- test if the connection to the clients are still working
- Deployed the CA file which has only the new Root CA on all clients
- Finally: update the CA Certificate in the Database

modules/administration/pages/custom-ssl.adoc
Loquacity: Yeah, I was wondering about that.

modules/administration/pages/custom-ssl.adoc
Loquacity: Is this written up somewhere?

mcalmer: Yes, on the same wiki page where you got the other content😉

Look for section "Replace the Server Certificate on the Server and the Proxies" up to the end of the page .https://wiki.microfocus.com/index.php/SUSE_Manager/Replace_CA_Certificate

Loquacity: :woman_facepalming:

modules/administration/pages/custom-ssl.adoc
```diff
@@ -194,3 +205,85 @@ Click btn:[Find Targets].
...
+rpm -Uhv ssl-build/susemanager-proxy/rhn-org-httpd-ssl-key-pair-susemanager-proxy-1.0-2.noarch.rpm
+----
+. Restart services to pick up the changes:
++
+----
+spacewalk-service restart
+----
+
+
+When you have replaced the server certificates on your server and any proxies, you need to update the certificate with only the new details all the clients.
```
keichwa: Maybe, "_on_ all the clients".

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,289 @@
...
 
+* A certificate authority (CA) SSL public certificate
+* An SSL server key
+* An SSL server certificate
 
+Your key and certificate files must be in PEM format.
 
-The following section will guide you through using a custom certificate with {productname} {productnumber} and {susemgrproxy} {productnumber}.
+The host name of the SSL keys and certificates must match the fully qualified host name of the machine you deploy them on.
+You can configure the host name of the certificate by editing the ``X509v3 Subject Alternative Name`` section of the certificate itself.
```
mcalmer:
```suggestion
You should set the host names in the ``X509v3 Subject Alternative Name`` section of the certificate.
```

"editing" sounds like you could use vi to change things, but certificates needs to be created using special tools.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,289 @@
...
+----
+// I would like to split up these steps, I think. LKB 2019-09-10
+. Deploy the CA certificate on the server:
++
 ----
+/usr/bin/rhn-deploy-ca-cert.pl --source-dir /root/combined-ssl-build \
+--target-dir /srv/www/htdocs/pub/ --trust-dir=/etc/pki/trust/anchors/
+----
+
+When you have the combined RPMs, you can deploy the combined certificates to your clients.
```
mcalmer:
```suggestion
When you have the combined RPMs, you can deploy the combined CA certificates to your clients.
```

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,289 @@
...
-Do you want to import existing certificates?
+rpm -Uhv ssl-build/susemanager-proxy/rhn-org-httpd-ssl-key-pair-susemanager-proxy-1.0-2.noarch.rpm
 ----
+. Restart services to pick up the changes:
 +
-Answer with kbd:[y]
-.
-. Continue by following the script prompts.
+----
+spacewalk-service restart
```
mcalmer:
```suggestion
rhn-proxy restart
```

For proxies we have a different command. And we do not have a symlink which name it mgr-proxy :-(

modules/administration/pages/custom-ssl.adoc
mcalmer: Maybe we should add here, that the customer should "test" if the clients still operate correctly and can reach the server/proxies via SSL.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,289 @@
...
+----
+rhnpush -c ssl-ca-channel --nosig \
+--ca-chain=/srv/www/htdocs/pub/RHN-ORG-TRUSTED-SSL-CERT \
+/root/ssl-build/rhn-org-trusted-ssl-cert-1.0-3.noarch.rpm
+----
+
+
+When you have the new certificate in the channel, you can use the {productname} {webui} to update it on all clients by synchronizing them with the channel.
+Alternatively, for Salt clients, you can use menu:Salt[Remote Commands].
+
```
mcalmer: Proxies are here also "just clients", but there is one special thing. In `/srv/www/htdocs/pub/` there is a copy of 
the CA Certificate and the rpm which contains it. The should be replaced as well. Proxies should have the same content as the server when it comes to certifcates:

```
RHN-ORG-TRUSTED-SSL-CERT
rhn-org-trusted-ssl-cert-*.noarch.rpm
```

modules/administration/pages/custom-ssl.adoc
mcalmer: I think we should change it and say 

. In the {productname} {webui} navigate to the Systems Overview Page and select all Salt Managed Clients into System Set Manager (SSM)
. Navigate to Systems => System Set Manager => Overview and select  `Apply system states`
. Apply the Highstate on all systems in the system set manager by clicking on the "Apply Highstate" button

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,289 @@
...
++
+----
+rhnpush -c ssl-ca-channel --nosig \
+--ca-chain=/srv/www/htdocs/pub/RHN-ORG-TRUSTED-SSL-CERT \
+/root/ssl-build/rhn-org-trusted-ssl-cert-1.0-3.noarch.rpm
+----
+
+
+When you have the new certificate in the channel, you can use the {productname} {webui} to update it on all clients by synchronizing them with the channel.
+Alternatively, for Salt clients, you can use menu:Salt[Remote Commands].
```
mcalmer: Maybe you can reference the "Deploying Combined Certificates on Salt Clients" section.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,296 @@
...
-. When prompted with:
+. On the {productname} Proxy, at the command prompt, install the RPM from the [path]``ssl-build`` directory:
 +
+----
+rpm -Uhv ssl-build/susemanager-proxy/rhn-org-httpd-ssl-key-pair-susemanager-proxy-1.0-2.noarch.rpm
+----
+. Restart services to pick up the changes:
++
+----
+rhn-proxy restart
```
mcalmer:
```suggestion
spacewalk-proxy restart
```

Just found out that we have a better alias which should be used here.

modules/administration/pages/custom-ssl.adoc
```diff
@@ -1,97 +1,296 @@
...
-
+. Install the {productname} Server according to the instructions in xref:installation:install-intro.adoc[].
+. Complete the initial setup according to xref:installation:server-setup.adoc[].
+. At the command prompt, point the SSL environment variables to the custom certificate file locations:
++
+----
+export CA_CERT=`path_to_CA_certificate_file`
+export SERVER_KEY=`path_to_web_server_key`
+export SERVER_CERT=`path_to_web_server_certificate`
+----
```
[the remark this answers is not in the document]
Loquacity: I've left the caps in, because the actual syntax requires the "quotes" and I didn't want to add more punctuation in. Replace the ``backticks`` with <pointy braces> though.

PR #702, opened 2019-09-09
Fix formatting issue

mcalmer:
Display correct list


modules/administration/pages/content-lifecycle.adoc
```diff
@@ -56,7 +56,8 @@ No package dependency solving is done on content filtering.
 === Filter ``rule`` Parameter
 Moreover, each filter has a ``rule`` parameter set to either ``Allow`` or ``Deny``.
 The filters are processed:
-- if a package or patch satisfies a ``Deny`` filter, it will be excluded from the result,
+
+* if a package or patch satisfies a ``Deny`` filter, it will be excluded from the result,
```
keichwa:
```suggestion
* If a package or patch satisfies a ``Deny`` filter, it will be excluded from the result.
```

PR #703, opened 2019-09-09
Add info about available Matcher algorithms

mcalmer:
We have multiple "Matcher" which should be explained.

### DIFF
--- a/modules/administration/pages/content-lifecycle.adoc
+++ b/modules/administration/pages/content-lifecycle.adoc
@@ -53,6 +53,17 @@ This is the list of supported filters:
 No package dependency solving is done on content filtering.
 ====
 
+Depending on the Filter Type multiple ``Matchers`` are available. There exists:
+
+* contains
+* matches (regexp)
+* equals
+* greater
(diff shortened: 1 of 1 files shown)


modules/administration/pages/content-lifecycle.adoc
```diff
@@ -53,6 +53,17 @@ This is the list of supported filters:
 No package dependency solving is done on content filtering.
 ====
 
+Depending on the Filter Type multiple ``Matchers`` are available. There exists:
```
Loquacity:
```suggestion
There are multiple matchers you can use with the filter.
Which ones are available will depend on the filter type you choose.
The full list is:
```

modules/administration/pages/content-lifecycle.adoc
```diff
@@ -53,6 +53,17 @@ This is the list of supported filters:
...
+
+* contains
+* matches (regexp)
```
Loquacity:
```suggestion
* matches (must take the form of a regular expression)
```

PR #705, opened 2019-09-10
use https://download.suse.com/index.jsp as default download URL

keichwa:
See also here: https://github.com/SUSE/spacewalk/issues/9040#issuecomment-520800508

### DIFF
--- a/modules/installation/pages/general-requirements.adoc
+++ b/modules/installation/pages/general-requirements.adoc
@@ -53,7 +53,7 @@ You only require a valid registration code for {productname}.
 You do not require a separate code for {slsa}{nbsp}{sles-version} {sp-version}.
 endif::[]
 
-If not already done, download the {sle} Unified Installer (version{nbsp}15{nbsp}SP1 or higher) from https://www.suse.com/products/server/download/[SUSE Linux Enterprise Server - Media Download].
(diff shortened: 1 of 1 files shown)


modules/installation/pages/general-requirements.adoc
```diff
@@ -53,7 +53,7 @@ You only require a valid registration code for {productname}.
 You do not require a separate code for {slsa}{nbsp}{sles-version} {sp-version}.
 endif::[]
 
-If not already done, download the {sle} Unified Installer (version{nbsp}15{nbsp}SP1 or higher) from https://www.suse.com/products/server/download/[SUSE Linux Enterprise Server - Media Download].
+If not already done, download the {sle} Unified Installer from https://download.suse.com/index.jsp, with the search parameters included it is one for example: https://download.suse.com/index.jsp?product_id=&search=Search&families=22609&version=68287 (version{nbsp}15{nbsp}SP1 or higher).
```
paususe: I'm not so sure about second part of the text ("with the search parameters included it is one for example")

What about something like this?

_If not already done, download the {sle} Unified Installer from https://download.suse.com/index.jsp . Direct link to {sle} 15 SP1, required to install SUSE Manager 4.0: https://download.suse.com/index.jsp?product_id=&search=Search&families=22609&version=68287_

keichwa: thanks, this flows much better :)

PR #713, opened 2019-09-17
Client Discovery by IP

Loquacity:
https://github.com/SUSE/spacewalk/issues/8223


Loquacity:
This works fine as concept content, but I think it would benefit from a task or two. I'm going to try and get some more detail from @joesusecom

Loquacity:
I've received no feedback, and this PR is rapidly aging. I'm going to assume there's no interest in this content for now. We can revisit if people ask.

PR #717, opened 2019-09-17
Update tuning-changelogs.adoc

juliogonzalez:
- Add a note about the need of restarting services after changing `/etc/rhn/rhn.conf`
- Improve the example of commands that the user needs to run:
  - Fix the prompts
  - Improve the style for SQL commands

### DIFF
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -31,7 +31,7 @@ yourdatabase=# DELETE FROM rhnPackageRepodata;
 yourdatabase=# INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
                (SELECT sequence_nextval('rhn_repo_regen_queue_id_seq'),
                        C.label,
-                       'package summary modification',
+                       'cached data regeneration',
                        'Y'
                 FROM rhnChannel C);
 yourdatabase=#  \q
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -14,16 +14,16 @@ java.max_changelog_entries = 0
 
 If you set this parameter (remember you need to restart the services with `spacewalk-service restart`), it will come into effect only for new packages when they are synchronized.
 
-You might like to delete the cached data to remove older data.
+You might like to delete and regenerate the cached data to remove older data.
 
 [IMPORTANT]
 ====
-Deleting cached data can take a long time.
+Deleting and regenerating cached data can take a long time.
 Depending on the number of channels you have and the amount of data to be deleted, it can potentially take several hours.
 The task is run in the background by Taskomatic, so you can continue to use {productname} while the operation completes, however you should expect some performance loss.
 ====
 
-You can delete cached data from the command line, using the SQL interface:
+You can delete and and request a regeneration of cached data from the command line, using the SQL interface:
 
 ----
 # spacewalk-sql -i
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -29,6 +29,9 @@ You can delete and request a regeneration of cached data from the command line:
 
 ----
 spacewalk-sql -i
+----
+Then on the SQL database prompt, enter:
+----
 DELETE FROM rhnPackageRepodata;
 INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
 (SELECT sequence_nextval('rhn_repo_regen_queue_id_seq'),
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -12,7 +12,7 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
 java.max_changelog_entries = 0
 ----
 
-If you set this parameter (remember you need to restart the services with `spacewalk-services restart), it will come into effect only for new packages when they are synchronized.
+If you set this parameter (remember you need to restart the services with `spacewalk-service restart`), it will come into effect only for new packages when they are synchronized.
 
 You might like to delete the cached data to remove older data.
 
@@ -28,11 +28,11 @@ You can delete cached data from the command line, using the SQL interface:
 ----
 # spacewalk-sql -i
 yourdatabase=# DELETE FROM rhnPackageRepodata;
-yourdatabase=#  INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
-(SELECT sequence_nextval('rhn_repo_regen_queue_id_seq'),
-        C.label,
-        'package summary modification',
-        'Y'
-   FROM rhnChannel C);
+yourdatabase=# INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
+               (SELECT sequence_nextval('rhn_repo_regen_queue_id_seq'),
+                       C.label,
+                       'package summary modification',
+                       'Y'
+                FROM rhnChannel C);
 yourdatabase=#  \q
 ----
--- a/modules/administration/pages/tuning-changelogs.adoc
+++ b/modules/administration/pages/tuning-changelogs.adoc
@@ -12,7 +12,8 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
 java.max_changelog_entries = 0
 ----
 
-If you set this parameter, it will come into effect only for new packages when they are synchronized.
+If you set this parameter (remember you need to restart the services with `spacewalk-services restart), it will come into effect only for new packages when they are synchronized.
+
 You might like to delete the cached data to remove older data.
 
 [IMPORTANT]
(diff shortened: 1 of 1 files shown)


modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,7 +12,8 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
 java.max_changelog_entries = 0
 ----
 
-If you set this parameter, it will come into effect only for new packages when they are synchronized.
+If you set this parameter (remember you need to restart the services with `spacewalk-service restart`), it will come into effect only for new packages when they are synchronized.
```
juliogonzalez: I am pretty sure we need the restart. @mcalmer?

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -25,13 +26,13 @@ The task is run in the background by Taskomatic, so you can continue to use {pro
...
+# spacewalk-sql -i
+yourdatabase=# DELETE FROM rhnPackageRepodata;
+yourdatabase=# INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
+               (SELECT sequence_nextval('rhn_repo_regen_queue_id_seq'),
+                       C.label,
+                       'package summary modification',
+                       'Y'
+                FROM rhnChannel C);
+yourdatabase=#  \q
 ----
```
juliogonzalez: I guess that what happens here is in fact:
1. We remove the cached data
2. We add all packages to a queue so the cached data is regenerated.

mcalmer: correct. I would also change 'package summary modification' into something more generic. AFAIK we copied this from a schema migration where the reason for re-generation was a change in the summary.

juliogonzalez: References changed to "delete and regenerate", but I am not sure what you mean about:

> I would also change 'package summary modification' into something more generic.

(maybe because I am not so familiar with the DB schema)

mcalmer: That string is just a text which should tell you why this job is there.
Not 100% sure if it is visible somewhere.

juliogonzalez: Ah!

juliogonzalez: Done.

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,26 +12,27 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
 java.max_changelog_entries = 0
 ----
 
-If you set this parameter, it will come into effect only for new packages when they are synchronized.
-You might like to delete the cached data to remove older data.
+If you set this parameter (remember you need to restart the services with `spacewalk-service restart`), it will come into effect only for new packages when they are synchronized.
+
+You might like to delete and regenerate the cached data to remove older data.
```
keichwa: Instead of "delete and regenerate" I'd use the term "purge".  Then in the IMPORTANT note, we can explain that purging in this context means deleting and then regenerating.

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,26 +12,27 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
...
 [IMPORTANT]
 ====
-Deleting cached data can take a long time.
+Deleting and regenerating cached data can take a long time.
 Depending on the number of channels you have and the amount of data to be deleted, it can potentially take several hours.
 The task is run in the background by Taskomatic, so you can continue to use {productname} while the operation completes, however you should expect some performance loss.
 ====
 
-You can delete cached data from the command line, using the SQL interface:
+You can delete and and request a regeneration of cached data from the command line, using the SQL interface:
```
keichwa:
```suggestion
You can delete and and request a regeneration of cached data from the command line:
```
Better avoid the confusing prompts (we have some variants in our docs...)

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,26 +12,27 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
...
-spacewalk-sql -i
-SQL> delete from rhnPackageRepodata;
-SQL> insert into rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
-(select sequence_nextval('rhn_repo_regen_queue_id_seq'),
-        C.label,
-        'package summary modification',
-        'Y'
-   from rhnChannel C);
-SQL> \q
+# spacewalk-sql -i
```
keichwa:
```suggestion
spacewalk-sql -i
----

Then, on the SQL database prompt:

```

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,26 +12,27 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
...
+yourdatabase=# DELETE FROM rhnPackageRepodata;
```
keichwa:
```suggestion
----
DELETE FROM rhnPackageRepodata;
```

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,26 +12,27 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
...
+yourdatabase=# INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
```
keichwa:
```suggestion
INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
```

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,26 +12,27 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
...
+               (SELECT sequence_nextval('rhn_repo_regen_queue_id_seq'),
+                       C.label,
+                       'cached data regeneration',
+                       'Y'
+                FROM rhnChannel C);
+yourdatabase=#  \q
```
keichwa:
```suggestion
\q
```

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -12,26 +12,27 @@ The parameter defaults to [systemitem]``0``, which means unlimited.
 java.max_changelog_entries = 0
 ----
 
-If you set this parameter, it will come into effect only for new packages when they are synchronized.
-You might like to delete the cached data to remove older data.
+If you set this parameter (remember you need to restart the services with `spacewalk-service restart`), it will come into effect only for new packages when they are synchronized.
```
Loquacity:
```suggestion
If you set this parameter, it will come into effect only for new packages when they are synchronized.
After changing this parameter, restart services with ``spacewalk-service restart``.
```

modules/administration/pages/tuning-changelogs.adoc
Loquacity: why?

juliogonzalez: @Loquacity: Why... what?

If it's about my change: because the change comes into effect only for new packages. For existing packages, "delete and regenerate" is needed so the old changelogs are removed from database, and then regenerated with the limits specified at the config file.

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -13,25 +13,28 @@ java.max_changelog_entries = 0
...
+INSERT INTO rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
+(SELECT sequence_nextval('rhn_repo_regen_queue_id_seq'),
         C.label,
-        'package summary modification',
+        'cached data regeneration',
         'Y'
-   from rhnChannel C);
-SQL> \q
+        FROM rhnChannel C);
+\q
```
juliogonzalez: Personally, I found this format somehow confusing, but I changed it as requested by @keichwa

modules/administration/pages/tuning-changelogs.adoc
Loquacity: I meant why use "purge"?

keichwa:
On 2019-09-18 09:44, Julio González Gil wrote:

> Personally, I found this format somehow confusing, but I changed it as
> requested by @keichwa

It was not meant as a strong request, but a suggestions ;)

Unfortunately we have different concepts which command prompt indicators
we want to use.  Or none at all.

(I'd enforce those things with appropriate markup and a good markup
language such as sgml, xml, texinfo, or latex.  Other template driven
tools (*office or *docs) are also fine.  Markup-less markup will not
help with those issues.

-- 
Karl Eichwalder

modules/administration/pages/tuning-changelogs.adoc
keichwa: That's what the user wants to do, IMO.  Person does not want to delete and regenerate (that's the command sequence to be run), but purging is actually wanted.  Maybe, there is a better term.

Of course, it is perfectly fine if we explain how this purging takes place technically.

I do not insist on my proposal.  Maybe, it causes more trouble than help.

modules/administration/pages/tuning-changelogs.adoc
```diff
@@ -13,25 +13,28 @@ java.max_changelog_entries = 0
...
 
-You can delete cached data from the command line, using the SQL interface:
+You can delete and request a regeneration of cached data from the command line:
 
 ----
 spacewalk-sql -i
-SQL> delete from rhnPackageRepodata;
-SQL> insert into rhnRepoRegenQueue (id, CHANNEL_LABEL, REASON, FORCE)
-(select sequence_nextval('rhn_repo_regen_queue_id_seq'),
+DELETE FROM rhnPackageRepodata;
```
keichwa:
```suggestion
----

Then on the SQL database prompt, enter:

----
DELETE FROM rhnPackageRepodata;
```
I think I failed to add this comment properly.  Hope this helps.

juliogonzalez: @keichwa done, I think. Please review :-)

modules/administration/pages/tuning-changelogs.adoc
juliogonzalez: With your new suggestion, I found it now easily understandable. Thanks a lot.

PR #720, opened 2019-09-17
Update clm doc

mcalmer:
We implemented some new filters and we had UI changes for the Filter setup.
This is now reflected in the docs as well


modules/administration/pages/content-lifecycle-examples.adoc
```diff
@@ -110,9 +114,9 @@ When you rebuild a previous environment with the btn:[Build] button, the new exc
 === Including a Patch in the Project
 
 In this example, you have received a security alert.
-An important security patch was released several days after your last release on 1 August.
+An important security patch was released several days after the first of the month you are currently working on.
```
mcalmer: We did not say in the beginning that we work in our example on the first of August. So I just say "this month".
If it is better to keep August 1st, then we should introduce it at the beginning where we explain what we want to do.

PR #721, opened 2019-09-17
minor update

mcalmer:
Just add the info what you need to run to get the sources synced.


modules/administration/pages/mirror-sources.adoc
```diff
@@ -25,6 +25,7 @@ spacewalk-service restart
 Currently, this feature can only be enabled globally for all repositories.
 It is not possible to select individual repositories for mirroring.
 
-When this feature has been activated, the source packages will become available in the {productname} {webui}.
+When this feature has been activated, the source packages will become available in the {productname} {webui}
+after the next repository sync.
```
Loquacity:
```suggestion
after the next repository synchronization.
```

PR #725, opened 2019-09-19
Disk Space Checker

Loquacity:
https://github.com/SUSE/spacewalk/issues/9457

### DIFF
--- a/modules/administration/pages/space-management.adoc
+++ b/modules/administration/pages/space-management.adoc
@@ -1,10 +1,9 @@
 [[space-management]]
 = Managing Disk Space
 
-If there is no available disk space, {productname} will not function.
-Recovering from this kind of error is extremely difficult.
+Running out of disk space can have a severe impact on the {productname} database and file structure which, in most cases, is not recoverable.
 
-{productname} monitors some directories for available disk space.
+{productname} monitors some directories for free disk space.
 You can modify which directories are monitored, and the warnings that are created.
 All settings are configured in the [path]``/etc/rhn/rhn.conf`` configuration file.
 
@@ -19,7 +18,7 @@ By default, {productname} monitors these directories:
 * [path]``/var/cache /srv``
 
 You can change which directories are monitored with the [systemitem]``spacecheck_dirs`` parameter.
-You can specify multiple directories by seaprating them with a space.
+You can specify multiple directories by separating them with a space.
 
 For example:
 
--- a/modules/administration/pages/space-management.adoc
+++ b/modules/administration/pages/space-management.adoc
@@ -1,7 +1,7 @@
 [[space-management]]
 = Managing Disk Space
 
-Running out of disk space can have a severe impact on the {productname} database and file structure which, in most cases, is not recoverable.
+Running out of disk space can have a severe impact on the {productname} database and file structure which, in some cases, is not recoverable.
 
 {productname} monitors some directories for free disk space.
 You can modify which directories are monitored, and the warnings that are created.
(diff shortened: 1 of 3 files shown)


modules/administration/pages/space-management.adoc
```diff
@@ -0,0 +1,68 @@
+[[space-management]]
+= Managing Disk Space
+
+If there is no available disk space, {productname} will not function.
```
keichwa: I think we must define what "available" means here.  "Enough disk space where {productname} can manage (write, store, delete, ...) data."  For this kind of oparation I used to use the term "free disk space".  Maybe, that's different now with all the new FSs such as btrfs.

Loquacity: reworded.

modules/administration/pages/space-management.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+----
+spacecheck_dirs = /var/lib/pgsql var/spacewalk /var/cache /srv
+----
+
+
+
+== Thresholds
+
+By default, {productname} will create a warning email when a monitored directory has less than 10% of total space available.
```
keichwa: Just spotting...  we use "email" and "e-mail" (the latter per SUSE style guide, iirc).  I'd rather vote to go for just "mail" :)

modules/administration/pages/space-management.adoc
```diff
@@ -0,0 +1,67 @@
...
+spacecheck_free_critical = 5
+----
+
+
+
+== Shut Down Services
+
+By default, {productname} will shut down the spacewalk services when the critical alert threshold is reached.
+
+You can disable this behaviour with the [systemitem]``spacecheck_shutdown`` parameter.
```
keichwa:
```suggestion
You can disable this behavior with the [systemitem]``spacecheck_shutdown`` parameter.
```
;)

modules/administration/pages/space-management.adoc
Loquacity: hrm, IBM says "email". I think I'm with you on erring on the side of "mail", though ;)

modules/administration/pages/space-management.adoc
Loquacity: *sigh*

modules/administration/pages/space-management.adoc
```diff
@@ -0,0 +1,67 @@
+[[space-management]]
+= Managing Disk Space
+
+Running out of disk space can have a severe impact on the {productname} database and file structure which, in most cases, is not recoverable.
```
mantel: Maybe "in some cases" instead of "most cases". I think postgresql also reserves some space to avoid running out of space, but it is good to make people aware that such a condition should be avoided by all means.

modules/administration/pages/space-management.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+
+
+== Monitored Directories
+
+By default, {productname} monitors these directories:
+
+* [path]``/var/lib/pgsql``
+* [path]``/var/spacewalk``
+* [path]``/var/cache /srv``
```
mantel: Those are two separate directories! /var/spacewalk and /srv

modules/administration/pages/space-management.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+== Shut Down Services
+
+By default, {productname} will shut down the spacewalk services when the critical alert threshold is reached.
+
+You can disable this behavior with the [systemitem]``spacecheck_shutdown`` parameter.
+
+For example:
+----
+spacecheck_shutdown = true
```
mantel: Maybe we could add a note that ANY value other than "true" will disable this.

modules/administration/pages/space-management.adoc
Loquacity: oops!

PR #735, opened 2019-09-19
Update icinga.adoc

juliogonzalez:
Related: https://github.com/SUSE/spacewalk/issues/9473

The uncommented part was reviewed and tested on real instances.

The commented part is, as discussed, full of Server references. If we ever want to publish it, we'll need to perform a cleanup and simplify that part, so there are no references to the server.

### DIFF
--- a/modules/administration/pages/icinga.adoc
+++ b/modules/administration/pages/icinga.adoc
@@ -137,7 +137,7 @@ vi /etc/icinga/ido2db.cfg
 [IMPORTANT]
 .Open Firewall Port
 ====
-Allow port `5432` through your firewall or you will not be able to access the WebGUI.
+Allow port `80` through your firewall or you will not be able to access the WebGUI.
 ====
 +
 
@@ -160,7 +160,7 @@ systemctl enable apache2.service
 systemctl start apache2.service
 ----
 
-. Login to the WebGUI at: http://localhost/icinga.
+. Login to the WebGUI at: http://<client-hostname>/icinga.
 
 This concludes setup and initial configuration of Icinga.
 
--- a/modules/administration/pages/icinga.adoc
+++ b/modules/administration/pages/icinga.adoc
@@ -137,7 +137,7 @@ vi /etc/icinga/ido2db.cfg
 [IMPORTANT]
 .Open Firewall Port
 ====
-Allow port `80` through your firewall or you will not be able to access the WebGUI.
+Allow port `80` through your firewall or you will not be able to access the {webui}.
 ====
 +
 
--- a/modules/administration/pages/icinga.adoc
+++ b/modules/administration/pages/icinga.adoc
@@ -35,7 +35,7 @@ SLES {sles-version} and later include these channels by default.
 +
 
 ----
-zypper in icinga icinga-idoutils-pgsql postgresql postgresql94-server \
+zypper in icinga icinga-idoutils-pgsql postgresql postgresql-server \
 monitoring-plugins-all apache2
 ----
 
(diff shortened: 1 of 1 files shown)


modules/administration/pages/icinga.adoc
```diff
@@ -128,7 +137,7 @@ vi /etc/icinga/ido2db.cfg
 [IMPORTANT]
 .Open Firewall Port
 ====
-Allow port `5432` through your firewall or you will not be able to access the WebGUI.
+Allow port `80` through your firewall or you will not be able to access the WebGUI.
```
juliogonzalez: Either I am quite wrong, or 5432 was incorrect. 5432 is for PostgreSQL, that it's to be used by ido2b when it receives data from nrpe.

modules/administration/pages/icinga.adoc
```diff
@@ -151,7 +160,7 @@ systemctl enable apache2.service
 systemctl start apache2.service
 ----
 
-. Login to the WebGUI at: http://localhost/icinga.
+. Login to the WebGUI at: http://<client-hostname>/icinga.
```
juliogonzalez: If we say 'localhost', the user needs to use a browser on the very same client. I think it makes more sense to use the hostname for the client.

modules/administration/pages/icinga.adoc
```diff
@@ -151,7 +160,7 @@ systemctl enable apache2.service
...
 
 This concludes setup and initial configuration of Icinga.
```
juliogonzalez: @Loquacity I am worried about the next section. It's about monitoring the server with Icinga, but AFAIK we don't support that anymore (we have prometheus for that).

Do we want to invest time on rewriting the whole section to describe how to monitor a simple version and then just point to the official doc?

Loquacity: I wouldn't bother personally. It seems pretty unlikely that this will be published any time soon. Just comment it out.

modules/administration/pages/icinga.adoc
```diff
@@ -57,19 +57,28 @@ define contact {
 systemctl enable postgresql.service
 systemctl start postgresql.service
 ----
-. Create the database and user for Icinga:
-+
-
+. Become postgres system user:
```
Loquacity:
```suggestion
. Become the ``postgres`` system user:
```

modules/administration/pages/icinga.adoc
```diff
@@ -57,19 +57,28 @@ define contact {
...
+----
+sudo su - postgres
+----
+. Start a SQL CLI:
+----
+psql
+----
+. Then setup a new password for the administrative user postgres, and configure a new icinga user:
```
Loquacity:
```suggestion
. Create a new password for the administrative user ``postgres``, and configure a new ``icinga`` user:
```

modules/administration/pages/icinga.adoc
```diff
@@ -57,19 +57,28 @@ define contact {
...
+----
+ALTER USER postgres WITH PASSWORD '<newpassword>';
+CREATE USER icinga;
+ALTER USER icinga WITH PASSWORD 'icinga';
+CREATE DATABASE icinga;
+GRANT ALL ON DATABASE icinga TO icinga;
+\q
+----
+Finally, exit the postgres system user session:
```
Loquacity:
```suggestion
Exit the ``postgres`` system user session:
```

modules/administration/pages/icinga.adoc
```diff
@@ -107,11 +116,11 @@ The entries in [path]``pg_hba.conf`` are read from top to bottom.
 systemctl reload postgresql.service
 ----
 
-. Configure the database schema by running the following command in [path]``/usr/share/doc/packages/icinga-idoutils-pgsql/pgsql/``:
+. Configure the database schema by running the following command:
```
Loquacity:
```suggestion
. Configure the database schema with this command:
```

modules/administration/pages/icinga.adoc
Loquacity:
```suggestion
Allow port `80` through your firewall or you will not be able to access the {webui}.
```

modules/administration/pages/icinga.adoc
```diff
@@ -151,12 +160,20 @@ systemctl enable apache2.service
 systemctl start apache2.service
 ----
 
-. Login to the WebGUI at: http://localhost/icinga.
+. Login to the WebGUI at: http://<client-hostname>/icinga.
```
Loquacity:
```suggestion
. Log in to the {webui} at: http://<client-hostname>/icinga.
```

modules/administration/pages/icinga.adoc
jcayouette: That is correct, thank you.

modules/administration/pages/icinga.adoc
jcayouette: This should probably just be *hostname*

modules/administration/pages/icinga.adoc
jcayouette: @juliogonzalez I agree with lana, lets comment this section out for now. We are promoting prometheus.

PR #745, opened 2019-09-23
Added terminal names content

Loquacity:
https://github.com/SUSE/spacewalk/issues/8843

### DIFF
--- a/modules/retail/pages/retail-terminal-names.adoc
+++ b/modules/retail/pages/retail-terminal-names.adoc
@@ -59,12 +59,12 @@ B002.terminal.example.com-c643
 ----
 
 
-The branch ID is the unique identifier for the branch server that the terminal is connected to.
+The [systemitem]``BranchID`` is the unique identifier for the branch server that the terminal is connected to.
 You can configure this value in the PXE formula settings for the branch server.
 
-The FQDN is the fully-qualified domain name of the terminal.
+The [systemitem]``FQDN`` is the fully-qualified domain name of the terminal.
 
-The unique ID is the first four characters of a generated machine identification number.
+The [systemitem]``UniqueID`` is the first four characters of a generated machine identification number.
 You can disable this behavior by toggling the [systemitem]``DISABLE_UNIQUE_SUFFIX`` parameter in the PXE formula settings.
 
 For more information about the PXE formula, see xref:salt:formula-pxe.adoc[].
@@ -101,11 +101,11 @@ For more information on the bind formula, see xref:salt:formula-bind.adoc[].
 . In the [guimenu]``IP Address`` field, type the static IP address you assigned to the terminal in the DHCP formula settings.
 . OPTIONAL: For multiple terminals, click btn:[Add Item] and fill in the details for each terminal.
 . Click btn:[Save Formula] to save the changes.
-. Apply the highstate on the branch server to apply the changes, and update the terminal with the new hostname.
+. Apply the highstate on the branch server to apply the changes.
 
 [[IMPORTANT]]
 ====
-If the terminal was previously registered using a name based on the hardware type instead of the hostname, you will need to delete the original registration.
+If the terminal was previously registered using a name based on the hardware type instead of the hostname, you will need to delete the previous registration.
 When you re-register the terminal, use the new terminal name.
 ====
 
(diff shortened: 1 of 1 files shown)


modules/retail/pages/retail-terminal-names.adoc
```diff
@@ -14,5 +30,115 @@ For example:
...
 ----
 
-You can adjust this behavior by toggling the [systemitem]``DISABLE_HOSTNAME_ID`` and [systemitem]``DISABLE_UNIQUE_SUFFIX`` parameters in the PXE formula settings.
-For more information about the PXE formula, see xref:retail:retail-formulas.adoc[].
+
+The branch ID is the unique identifier for the branch server that the terminal is connected to.
+You can configure this value in the PXE formula settings for the branch server.
+
+The manufacturer, product name, and serial number are provided by the terminal hardware BIOS.
+If the terminal does not provide a serial number, it will be replaced in the terminal name with a hyphen (``-``).
```
aaannz: This is not entirely correct. It the terminal hardware does not provide serial number, then serial number will be omitted.

Loquacity:
Thanks @aaannz :smile: 
@keichwa can I get a review, please?

modules/retail/pages/retail-terminal-names.adoc
```diff
@@ -14,5 +30,115 @@ For example:
...
+For more information on the bind formula, see xref:salt:formula-bind.adoc[].
+. In the [guimenu]``Hostname`` field, type the hostname of the branch server.
+. In the [guimenu]``IP Address`` field, type the static IP address you assigned to the terminal in the DHCP formula settings.
+. OPTIONAL: For multiple terminals, click btn:[Add Item] and fill in the details for each terminal.
+. Click btn:[Save Formula] to save the changes.
+. Apply the highstate on the branch server to apply the changes, and update the terminal with the new hostname.
+
+[[IMPORTANT]]
+====
+If the terminal was previously registered using a name based on the hardware type instead of the hostname, you will need to delete the original registration.
```
keichwa: Instead of "original" I'd rather repeat the previously use word.  Something like: "to delete the previous registration" or "this previous ...".

modules/retail/pages/retail-terminal-names.adoc
```diff
@@ -14,5 +30,115 @@ For example:
...
+        IP: <IP_Address>
+        hwAddress: <MAC_Address>
+        hwtype: <HWTYPE_Group_name_without_HWTYPE:_prefix>
+----
+. Import the modified YAML file:
++
+----
+retail_yaml --from-yaml retail.yaml
+----
+. Apply the highstate on the branch server to apply the changes, and update the terminal with the new hostname.
```
keichwa: Maybe, the comma is superfluous or we could say something like this:
```suggestion
. Apply the highstate on the branch server to apply the changes, and thus updating the terminal with the new hostname.
```
Otherwise I read it as if "update the terminal with the new hostname" would be a second step.  But my impression could be wrong, misguided by the way we use commas in German.

modules/retail/pages/retail-terminal-names.adoc
```diff
@@ -14,5 +30,115 @@ For example:
...
+
+[[IMPORTANT]]
+====
+If the terminal was previously registered using a name based on the hardware type instead of the hostname, you will need to delete the original registration.
```
keichwa: Again the thing with the "original registration".

modules/retail/pages/retail-terminal-names.adoc
Loquacity: Yeah, that's a serial (Oxford) comma, so the part of the sentence after the comma is not a separate clause. I think removing the comma might make it clearer, but also not en-US ;)

Loquacity: Looking at this closer, that last part is a result statement anyway, so deleting it is probably the best thing to do. Let me know what you think.

Loquacity:
> Please decide yourself. My ideas could be wrong that early in the morning ;)
> 
> In general, I'd re-use the names of the replaceable strings literally (as `BranchID` or ). But in this case the human readable variants such as "branch ID" are probably clear enough.

Actually, I agree with you here. It makes more like a definition list, which should make it clearer that these things are defined further down the page. Good thinking :)

PR #747, opened 2019-09-23
Update clients-rh.adoc

juliogonzalez:
- "Set up a Red Hat Enterprise Linux Client" was incorrect. That part described stuff about the server
- First we need to sync the product from SCC, so we have the base channel, then the media is imported.
- Separate the registration to its own section (I removed the commented content, which was about this, and it's now a link).

Addresses:
- [ ] https://github.com/SUSE/spacewalk/issues/9347
- [ ] https://github.com/SUSE/spacewalk/issues/9529

### DIFF
--- a/modules/client-configuration/pages/clients-rh.adoc
+++ b/modules/client-configuration/pages/clients-rh.adoc
@@ -6,25 +6,7 @@ This section contains information about registering traditional and Salt clients
 
 
 
-== Set up a {rhel} Client
-
-Ensure that your client meets these requirements before you start:
-
-* 8{nbsp}GB RAM or more
-* Two or more physical or virtual CPU cores
-* Access to {rhel} installation and subscription media
-* An LVM or an NFS mount is recommended
-
-You will need to ensure you provision enough disk space.
-The [path]``/var/spacewalk`` directory contains all mirrored RPMs,  and can take a significant amount of disk space.
-For example, the {rhel}{nbsp}6 x86_64 channels require over 90{nbsp}GB.
-
-Taskomatic will use one CPU core, and requires at least 3072{nbsp}MB RAM.
-To ensure that Taskomatic has access to enough memory, open the [path]``/etc/rhn/rhn.conf`` configuration file, and add this line:
-
-----
-taskomatic.java.maxmemory=3072
-----
+== Server requirements
 
 [WARNING]
 ====
@@ -33,7 +15,20 @@ You must obtain support from either Red Hat or {suse} for all your RHEL systems.
 If you do not do this, you might be violating your terms with Red Hat.
 ====
 
+Ensure that your server meets the requirements outlined at installation:hardware-requirements.adoc
 
+Taskomatic will use one CPU core, and requires at least 3072{nbsp}MB RAM.
+
+To ensure that Taskomatic has access to enough memory, open the [path]``/etc/rhn/rhn.conf`` configuration file, add this line:
+
+----
+taskomatic.java.maxmemory=3072
+----
+
+And restart Taskomatic:
+----
+systemctl restart taskomatic
+----
 
 == {rhel} Channel Management
 
--- a/modules/client-configuration/pages/clients-rh.adoc
+++ b/modules/client-configuration/pages/clients-rh.adoc
@@ -1,19 +1,17 @@
 [[clients-rh]]
 = Registering {rhel} Clients
 
-
 This section contains information about registering traditional and Salt clients running {rhel} operating systems.
-
-
-
-== Server requirements
 
 [WARNING]
 ====
-You are responsible for arranging access to Red Hat base media repositories and RHEL installation media.
-You must obtain support from either Red Hat or {suse} for all your RHEL systems.
-If you do not do this, you might be violating your terms with Red Hat.
+{rhel} Clients are completely based on {redhat} and unrelated to {sleses} Clients (also known as RES or {redhat} {slesa})
+You are responsible for arranging access to {redhat} base media repositories and {rhela} installation media, as well as connecting {productname} Server to {redhat} Content Delivery Network
+You must obtain support from either {redhat} all your {rhela} systems.
+If you do not do this, you might be violating your terms with {redhat}.
 ====
+
+== Server requirements
 
 Ensure that your {productname} Server meets the requirements outlined at xref:installation:hardware-requirements.adoc[].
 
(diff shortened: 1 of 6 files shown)


modules/client-configuration/pages/clients-rh.adoc
```diff
@@ -6,36 +6,67 @@ This section contains information about registering traditional and Salt clients
...
 
-You will need to ensure you provision enough disk space.
-The [path]``/var/spacewalk`` directory contains all mirrored RPMs,  and can take a significant amount of disk space.
-For example, the {rhel}{nbsp}6 x86_64 channels require over 90{nbsp}GB.
+Ensure that your {productname} Server meets the requirements outlined at installation:hardware-requirements.adoc
 
 Taskomatic will use one CPU core, and requires at least 3072{nbsp}MB RAM.
-To ensure that Taskomatic has access to enough memory, open the [path]``/etc/rhn/rhn.conf`` configuration file, and add this line:
+
+To ensure that Taskomatic has access to enough memory, open the [path]``/etc/rhn/rhn.conf`` configuration file, add this line:
```
juliogonzalez: @mcalmer is it just me, or this is incorrect after the discussion I saw at the mailing list?

I guess user should not edit `/usr/share/rhn/config-defaults/rhn_taskomatic_daemon.conf` as that would get overwritten by an update.

What should we do?

modules/client-configuration/pages/clients-rh.adoc
```diff
@@ -6,36 +6,67 @@ This section contains information about registering traditional and Salt clients
...
-Ensure that your client meets these requirements before you start:
-
-* 8{nbsp}GB RAM or more
-* Two or more physical or virtual CPU cores
-* Access to {rhel} installation and subscription media
-* An LVM or an NFS mount is recommended
+[WARNING]
+====
+You are responsible for arranging access to Red Hat base media repositories and RHEL installation media.
+You must obtain support from either Red Hat or {suse} for all your RHEL systems.
```
juliogonzalez: `or`? Shouldn't it be `and` as we request the user to use an original media, and then use RES?

modules/client-configuration/pages/clients-rh.adoc
```diff
@@ -6,36 +6,67 @@ This section contains information about registering traditional and Salt clients
...
+If you do not do this, you might be violating your terms with Red Hat.
+====
 
-You will need to ensure you provision enough disk space.
-The [path]``/var/spacewalk`` directory contains all mirrored RPMs,  and can take a significant amount of disk space.
-For example, the {rhel}{nbsp}6 x86_64 channels require over 90{nbsp}GB.
+Ensure that your {productname} Server meets the requirements outlined at installation:hardware-requirements.adoc
```
Loquacity:
```suggestion
Ensure that your {productname} Server meets the requirements outlined at xref:installation:hardware-requirements.adoc[].
```

modules/client-configuration/pages/clients-rh.adoc
Loquacity: Seems so. Maybe say "You must obtain support from both Red Hat and {suse} for all your RHEL systems.:"

modules/client-configuration/pages/clients-rh.adoc
```diff
@@ -6,36 +6,67 @@ This section contains information about registering traditional and Salt clients
...
 taskomatic.java.maxmemory=3072
 ----
 
-[WARNING]
-====
-You are responsible for arranging access to Red Hat base media repositories and RHEL installation media.
-You must obtain support from either Red Hat or {suse} for all your RHEL systems.
-If you do not do this, you might be violating your terms with Red Hat.
-====
+And restart Taskomatic:
```
Loquacity:
```suggestion
Restart taskomatic:
```

modules/client-configuration/pages/clients-rh.adoc
```diff
@@ -6,36 +6,67 @@ This section contains information about registering traditional and Salt clients
...
+
+
+{rhel} 6::
+* Product: {rhel} 6
+* Mandatory channels: [systemitem]``rhel-x86_64-server-6`` , [systemitem]``res6-suse-manager-tools-x86_64`` , [systemitem]``res6-x86_64``
+
+
+There are two ways to check if a channel has finished synchronizing:
+
+* In the {productname}{webui}, navigate to menu:Admin[Setup Wizard] and select the menu:SUSE Products[] tab.
```
Loquacity:
```suggestion
* In the {productname} {webui}, navigate to menu:Admin[Setup Wizard] and select the menu:SUSE Products[] tab.
```

modules/client-configuration/pages/clients-rh.adoc
```diff
@@ -6,36 +6,67 @@ This section contains information about registering traditional and Salt clients
...
+You will need to check all the base and child channel log files to be sure that the synchronization is complete.
+
+[IMPORTANT]
+====
+Missing packages will cause your registration to fail.
+For {rhel} clients, packages are contained on the {rhel} installation media.
+Ensure you have loop-mounted the installation media, and added it as a child channel to the base {rhel} channel.
+====
+
+=== Adding {rhel} media
```
Loquacity:
```suggestion
=== Add {rhel} Media
```

modules/client-configuration/pages/clients-rh.adoc
mcalmer: I am not sure if `and` is correct. What do you want to say ?

modules/client-configuration/pages/clients-rh.adoc
mcalmer: Yes, but the text say rhn.conf which is ok. But i need to check the code again. I think we have a new configfile which should be used. We keep rhn.conf only for memory as backward compatible solution.

mantel: /etc/rhn/rhn.conf is correct for setting the taskomatic max memory.

modules/client-configuration/pages/clients-rh.adoc
juliogonzalez: @mcalmer, that you need support from Red Hat (to be able to get RHEL installation media and have the right to use it) `AND` SUSE (to be able to get RES updates).

Isn't that the case?

modules/client-configuration/pages/clients-rh.adoc
juliogonzalez: Done.

modules/client-configuration/pages/clients-rh.adoc
juliogonzalez: Done.

modules/client-configuration/pages/clients-rh.adoc
juliogonzalez: Done.

modules/client-configuration/pages/clients-rh.adoc
mcalmer: No. If People have official RHEL contract, they would not buy RES. RES is there to not need a subscription from RedHat.

RES was for "switching" to SUSE. The idea:

- You had a subscription with RedHat in the past
- From that times you have a media
- Now you switch to RES and use it for updates only.
- When you install a new server you use SLES and not RHEL.

(But this is theory, in practice people do not switch. But this is there problem)

Btw: A valid media os also "CentOS" for RES. And this is "free".

modules/client-configuration/pages/clients-rh.adoc
juliogonzalez: Done

modules/client-configuration/pages/clients-rh.adoc
juliogonzalez: I understand that, but then:
> You must obtain support from either Red Hat or {suse} for all your RHEL systems.

Why "Red Hat **OR** SUSE"? With your explanation, I'd say this warning is not needed.

> Btw: A valid media os also "CentOS" for RES. And this is "free".

Is this supported? If so, it should be mentioned at this manual, I guess.

juliogonzalez:
Superseeded by https://github.com/SUSE/doc-susemanager/pull/759

PR #756, opened 2019-09-25
Update proxy-migration.adoc

keichwa:


keichwa:
@mantel I'd like to combine your both proxy-migration.adoc patches.  (I hope I do this correctly with git and github).

PR #761, opened 2019-09-25
Update network-requirements.adoc

mantel:
Current wording sounds as if a not correctly set up DNS only would lead to some minor annoyances. In fact, this can result in a lot of severe problems up to complete loss of functionality. But the final wording of course should be done by a native English speaker.

### DIFF
--- a/modules/installation/pages/network-requirements.adoc
+++ b/modules/installation/pages/network-requirements.adoc
@@ -4,7 +4,7 @@
 This section details the networking and port requirements for {productname}.
 
 Fully Qualified Domain Name (FQDN)::
-The {productname} server must resolve its FQDN correctly or cookies will not work properly on the WebUI.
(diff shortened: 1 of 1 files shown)


modules/installation/pages/network-requirements.adoc
```diff
@@ -4,7 +4,7 @@
 This section details the networking and port requirements for {productname}.
 
 Fully Qualified Domain Name (FQDN)::
-The {productname} server must resolve its FQDN correctly or cookies will not work properly on the WebUI.
+The {productname} server must resolve its FQDN correctly. This is a hard requirement else various components might malfunction.
```
Loquacity:
```suggestion
The {productname} server must resolve its FQDN correctly. 
If the FQDN cannot be resolved, it can cause serious problems in a number of different components.
```

PR #764, opened 2019-09-25
Update install-vm.adoc

mantel:
Even for a test server we require 8 GB of RAM and 4 CPUs. As this section references the official hardware requirements, it should be consistent.

### DIFF
--- a/modules/installation/pages/install-vm.adoc
+++ b/modules/installation/pages/install-vm.adoc
@@ -30,8 +30,8 @@ Find the JeOS image at https://download.suse.com/.
 | Installation Method | Import Existing Disk Image
 | OS:                 | Linux
 | Version:            | SLES_version_-JeOS-for-kvm-and-xen.x86_64-GM.qcow2
-| Memory:             | 4096 MB
-| CPU's:              | 2
+| Memory:             | 8192 MB
+| CPU's:              | 4
 | Storage Format:     | _.qcow2_ 24 GB (Default) JeOS Root Partition
 | Virtual Disks:      |
 | VirtIO Disk 2       | 101 GB for [path]``/var/spacewalk``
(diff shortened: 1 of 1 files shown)


keichwa:
Hubert Mantel <notifications@github.com> writes:

> Even for a test server we require 8 GB of RAM and 4 CPUs. As this
> section references the official hardware requirements, it should be
> consistent.

Is this a hard requirement?  Last time I tried there was a warning but
you could install and run SUMA with 1 SLE client with reduced
ressources.  Because this is a VM you can add more ressources without
too much trouble.

But all this depends on what PMs wants this installation scenario for.
In the past, Joe wanted something as small as possible for a quick demo
or test.  This might have changed in the meantime, and consistency is
more important.

Thus I also approve your edits.  I probably should have added these
considerations as a comment to the text.

-- 
                           90409 Nürnberg, Germany
Geschäftsführer: Felix Imendörffer (HRB 247165, AG München)

mantel:
> Hubert Mantel <notifications@github.com> writes:
> Even for a test server we require 8 GB of RAM and 4 CPUs. As this section references the official hardware requirements, it should be consistent.
> Is this a hard requirement? Last time I tried there was a warning but you could install and run SUMA with 1 SLE client with reduced ressources. Because this is a VM you can add more ressources without too much trouble. But all this depends on what PMs wants this installation scenario for. In the past, Joe wanted something as small as possible for a quick demo or test. This might have changed in the meantime, and consistency is more important. Thus I also approve your edits. I probably should have added these considerations as a comment to the text.
> […](#)
> -- Karl Eichwalder SUSE Software Solutions Germany GmbH PES / Documentation Maxfeldstraße 5 90409 Nürnberg, Germany Geschäftsführer: Felix Imendörffer (HRB 247165, AG München)

I do not know if this is a hard requirement. Of course I know that you are able to install the system on a machine with less resources. I just caught my attention because it was using the term "test installation".

keichwa:
ok, then we'd better increase the values to avoid further confusion.

PR #765, opened 2019-09-25
Update install-ibmz.adoc

mantel:
Some smaller issues, but the part about actually installing the system might be incorrect. Please note the "TBD:"

### DIFF
--- a/modules/installation/pages/install-ibmz.adoc
+++ b/modules/installation/pages/install-ibmz.adoc
@@ -90,7 +90,7 @@ The hostname should contain less than eight characters and must not contain any
 
 This section covers the installation of {productname}{nbsp}{productnumber} as an extension to {sls}{nbsp}{sles-version}{nbsp}{sp-version}.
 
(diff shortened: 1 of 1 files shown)


modules/installation/pages/install-ibmz.adoc
```diff
@@ -90,6 +90,8 @@ The hostname should contain less than eight characters.
 
 This section covers the installation of {productname}{nbsp}{productnumber} as an extension to {sls}{nbsp}{sles-version}{nbsp}{sp-version}.
 
+TBD: Is this correct? SUSE Manager as an extension? Doesn't SLES15SP1 on z also have the new module concept with SUSE Manager being a base product?
```
Loquacity:
```suggestion
// TBD: Is this correct? SUSE Manager as an extension? Doesn't SLES15SP1 on z also have the new module concept with SUSE Manager being a base product?
```

Putting this in a comment to avoid accidents :wink:

PR #768, opened 2019-09-26
Retail: new arch section & network diagrams

Loquacity:
https://github.com/SUSE/spacewalk/issues/6051

There's probably lots more to be done here, and I would start with this: https://documentation.suse.com/sle-pos/11-SP3/html/SLEPOS-guide/cha.slepos_overview.html#sec.overview.architecture

(Anyone willing to prettify my diagrams?)


Loquacity:
Screenshot from 2019-09-26 14-52-53

modules/retail/pages/retail-components.adoc
```diff
@@ -1,13 +1,8 @@
...
 
-{smr} uses a layered architecture:
-
-* The first layer contains the {susemgr} server and one or more build hosts.
-* The second layer contains one or more branch servers to provide local network and boot services.
-* The final layer contains any number of deployed point-of-service terminals.
-
-image::retail_ref_arch.png[scaledwidth=80%]
+{smr} is made up of various components.
+For more on these components work together, see xref:modules/retail/pages/retail-network-arch.adoc[].
```
keichwa: Is this a complete sentence? At least, it's rather dense.  Better inserting "how" after "on"?

aaannz:
I'm sorry I forgot to provide you the source diagrams and those you have here are not correct. In the architecture overview, build host are not between branch and manager, they are separately of the retail chain and usually even one is sufficient.
Architecture looks like this
architecture

Then networking. Dedicated networking uses dedicated network for terminals and looks like this:
networking

Shared network on the other hand does not have this dedicated network and looks like this:
networking-shared

Loquacity:
image

Bischoff:
Thanks for all the hard work @Loquacity .

aaannz:
I think the previous images were more readable - when the names were next to the icons and not over them. Also maybe instead of 'Router' you can use 'Corporate WAN'.

Next, when I read the text I would prefer to have some clear separation between `Dedicated networks mode` mode and `Shared network mode`. Currently I would be confused where starts one and where the other.

Also the first sentence of the shared network I think can be improved:
'The branch server and the terminals are connected to the same network as the SUSE Manager server.'

Loquacity:
> I think the previous images were more readable - when the names were next to the icons and not over them. Also maybe instead of 'Router' you can use 'Corporate WAN'.

My diagram skills are not great :worried:  I will try to get back around to this. 

In the meantime, I've fixed the wording, though.

Loquacity:
I'm going to go ahead and merge, in the knowledge that the images will need more prettiness.

PR #775, opened 2019-09-30
Added admon about most recent docs

Loquacity:
https://github.com/SUSE/spacewalk/issues/9584

### DIFF
--- a/modules/ROOT/pages/index.adoc
+++ b/modules/ROOT/pages/index.adoc
@@ -21,6 +21,12 @@ ifeval::[{suma-content} == true]
 
 The following documentation is available for {productname} version {productnumber}.
 
+[IMPORTANT]
+====
+{productnamee} documentation is available in several locations and formats.
+For the most up-to-date version of this documentation, check https://opensource.suse.com/doc-susemanager first.
+====
+
 Download All PDFs icon:caret-right[] icon:file-archive[link="../susemanager-docs_en-pdf.tar.gz"]
 
 
@@ -65,7 +71,3 @@ Download All PDFs icon:caret-right[] icon:file-archive[link="../uyuni-docs_en-pd
 //| Architecture               | xref:architecture:architecture-intro.adoc[HTML] link:../pdf/suse_manager_architecture.pdf[PDF]
 |===
 endif::[]
-
-
-
-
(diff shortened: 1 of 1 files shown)


paususe:
@keichwa is right, I didn't mean to add opensource.suse.com but documentation.suse.com

Loquacity:
> The "_For the most up-to-date version of this documentation_" makes me a bit uncomfortable, it may lead a 4.0.2 customer to think he'll find the 4.0.2 docs at https://documentation.suse.com/suma/4.0/ but what he's going to find there is maybe 4.0.4
> 
> Maybe rephrase as "_For the most up-to-date version of the SUSE Manager 4.0 documentation..._" ?

I don't understand. Even if a customer is using 4.0.2, the *latest* documentation is still 4.0.4 (or whatever).

keichwa:
On 2019-10-02 03:31, Lana Brindley wrote:

> I don't understand. Even if a customer is using 4.0.2, the *latest*
> documentation is still 4.0.4 (or whatever).

Yes, but all this is not that east and self-evident.  To avoid confusion
it would make sense if we'd always introduce new feature in our docs 
with
a statement such as

     This feature or change is available since version x.y.z.

If the user is still on x.y.(z-1), person knows not to be affected.

That's a lot of work and often not easy to get it right.  Then, later 
on,
we must remove these version info.

We should check whether RH still provides documentation for every minor
release...

-- 
Karl Eichwalder

Loquacity:
> On 2019-10-02 03:31, Lana Brindley wrote: I don't understand. Even if a customer is using 4.0.2, the *latest* documentation is still 4.0.4 (or whatever).
> Yes, but all this is not that east and self-evident. To avoid confusion it would make sense if we'd always introduce new feature in our docs with a statement such as This feature or change is available since version x.y.z. If the user is still on x.y.(z-1), person knows not to be affected. That's a lot of work and often not easy to get it right. Then, later on, we must remove these version info. 

Yeah, this is because we haven't done versioning very well. Also, we really shouldn't add features in minor releases, but that's outside of my remit ;)

> We should check whether RH still provides documentation for every minor release...

When I was there, policy was x.y releases only, no x.y.z releases.

Loquacity:
@paususe Can you please approve so we can merge this?

PR #777, opened 2019-09-30
fix faulty merge

keichwa:
add forgotten changes and remove proxy-upgrade.adoc

(don't know why but I forgot these changes...)


keichwa:
Thanks for checking, @juliogonzalez !

PR #759, opened 2019-09-25
WIP: Rework ES/RHEL/CentOS doc

juliogonzalez:
General rework. Among other things:
- New `clients-sleses.adoc` for SUSE Linux Enterprise Server with Expanded Support (AKA RES, AKA Red Hat Expanded Support).
- `clients-rh.adoc` now discribes the procedure for real RHEL clients, with including how to connect the server to the CDN. The clients to be used are those for Expanded Support Client Tools for SUSE Manager, or Uyuni Client Tools for Uyuni (uses an `if`)
- `clients-centos.adoc` now describes how to onboard CentOS with Expanded Support Client Tools for SUSE Manager, or with Uyuni Client Tools for Uyuni (uses an `if`)

**IMPORTANT**: The CentOS part is NOT valid for 3.2 unless we backport changes for spacewalk-common-channels there.

I added `if` where needed so we can show the correct warnings for either SUSE Manager on Uyuni.

This PR will require adding the screenshots where I left comments, and for sure it will require checking for small typos, styling, etc :-)

For minor stuff, please do it directly at the PR.

If you have questions, please ask before changing :-)

**WARNING:** To review the `.adoc` files, I suggest you check the whole file, and not the diffs (go to 'Files changed' tab, then click `...` on the right side of the bar where the filename is, then `View file`)

Addresses:
- [ ] SUSE/spacewalk#9347

PR #778, opened 2019-09-30
typo

asemen:
should be centos7-uyuni-client not centos7-uyuni-client-x86_64


Loquacity:
This file is being re-written in https://github.com/SUSE/doc-susemanager/pull/759 and I think that might conflict with this PR.

PR #780, opened 2019-10-01
Added saltboot to components, add image file

Loquacity:
https://github.com/SUSE/spacewalk/issues/8857


Loquacity:
image

Loquacity:
Note: I have not linked the diagram file into the nav, because it's really only useful in the context of the associated words, but the image is too large to include in the components section.

modules/retail/pages/retail-components.adoc
```diff
@@ -93,7 +87,16 @@ Then all devices of a particular type can be managed as one.
 
 
 
-[[retail.sect.components.fitting.formulas]]
+=== System groups
```
keichwa:
```suggestion
=== System Groups
```

modules/retail/pages/retail-components.adoc
```diff
@@ -104,15 +107,33 @@ You can use formulas to apply configuration patterns to your hardware groups.
...
+Saltboot is a collection of tools and processes that are used to bootstrap, deploy and validate {smr} terminals.
 
-Each branch requires a system group, containing a single branch server, and the PoS terminals associated with that server.
-Each system group is identified with a Branch ID.
-The Branch ID is used in formulas and scripts to automatically update the entire group.
+Saltboot consists of:
+
+* Initialization service:
++
+The Saltboot ``initrd`` service is created during image building and is required for bootstrapping terminals.
```
aaannz: I wouldn't call `initrd` a service. It is a bunch of scripts and tools which are run right after kernel is loaded and are meant to prepare the machine to some state before system image takes over. (see https://en.wikipedia.org/wiki/Initial_ramdisk ).
I would really just call it `initrd`, in this case `saltboot initrd`.

modules/retail/pages/retail-components.adoc
```diff
@@ -104,15 +107,33 @@ You can use formulas to apply configuration patterns to your hardware groups.
...
++
+The Salt state that contains the logic for the entire saltboot process.
+
+* Partitioning pillars:
++
+The Salt pillar structure that describes how terminals are partitioned and what image is deployed on each terminal.
+
+* Images and bootimages pillars:
++
+These Salt pillars are generated after the Saltboot ``initrd`` service has successfully built an image.
```
aaannz: This is not correct. Should be
> These Salt pillars are generated after image with Saltboot ``initrd`` has been successfully built.

initrd is not building image, but part of the built image is initrd.

modules/retail/pages/retail-components.adoc
aaannz: Maybe even call it Branch System Groups.
Because hardware types also uses system groups, but they are mentioned above and this paragraph is solely about system groups named after branch id and used for per-branch management.

modules/retail/pages/retail-components.adoc
Loquacity: huh, i always thought the d was for daemon, didn't realise it was RAM disk! That makes a lot more sense now :wink:

modules/retail/pages/retail-components.adoc
```diff
@@ -104,15 +107,33 @@ You can use formulas to apply configuration patterns to your hardware groups.
...
++
+The Salt state that contains the logic for the entire saltboot process.
+
+* Partitioning pillars:
++
+The Salt pillar structure that describes how terminals are partitioned and what image is deployed on each terminal.
+
+* Images and boot images pillars:
++
+These Salt pillars are generated after saltboot has successfully built an image.
```
aaannz: One last correction. Saltboot is not building the image. Image is build using SUMA image building. The difference between other images (e.g. for virtualization) and retail is that the retail images contain saltboot initrd.
SUMA image building detects that and when the image does have saltboot initrd it generates these salt pillars.

So the sentence should be hinting that these salt pillars are generated after image with saltboot initrd has been successfuly built.

Loquacity: OK, I've tried again! Am I getting closer?

PR #760, opened 2019-09-25
TR feedback by moio

keichwa:
https://github.com/SUSE/spacewalk/issues/9512
Broken PR, to be closed...

PR #781, opened 2019-10-01
create clean maintenance window diff

keichwa:
addressing TR feedback by moio (SUSE/spacewalk#9512)
close SUSE/spacewalk#9512
close https://github.com/SUSE/doc-susemanager/pull/760

### DIFF
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -13,7 +13,7 @@ You must keep it up-to-date or you will risk not being able to update something 
 
 == {productname} Server
 
-This section contains a checklist for your maintenance window, with links to further information on performing each of the steps, if you need it.
+This section contains a checklist for your maintenance window, with links to further information on performing each of the steps.
 
 // ke, 2019-09-30: we'll stop spacewalk during the update
 // . Stop spacewalk services.
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -20,7 +20,7 @@ This section contains a checklist for your maintenance window, with links to fur
 // . Check if the configuration is still correct.
 . Apply the latest updates.
 See xref:upgrade:server-update.adoc[].
-. Update to the latest service pack, if required.
+. Upgrade to the latest service pack, if required.
 See xref:upgrade:migrate-4x-4x.adoc[].
 // We reboot during the above listed procedures.
 // . Reboot the server.
@@ -37,7 +37,8 @@ For information on using zypper, see https://www.suse.com/documentation/sles-15/
 ////
 Preferable, you will run such a tool within a maintenance window; for more information, see xref:administration:maintenance-window.adoc#maintenance-window[].
 
- complete procedure, also see above:
+complete procedure, also see above:
+
 1. Log in as root user to the SUSE Manager server.
 2. Stop the Spacewalk service:
    spacewalk-service stop
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -99,11 +99,11 @@ This procedure is also known as a Service Pack (SP) migration.
 === Refreshing Client Tools After Server Update
 
 When the server is updated consider to update some tools on the clients, too.
-Updating salt-minion, zypper and any other related management package on clients is not a strict requirement, but it is a best practice in general.
-For example, a maintenance update on the Server might introduce a major new Salt version.
+Updating [package]``salt-minion``, [package]``zypper``, and other related management package on clients is not a strict requirement, but it is a best practice in general.
+For example, a maintenance update on the server might introduce a major new Salt version.
 Then minions will continue to work but might experience problems later on.
 To avoid this always update the salt-minion package when available.
-{suse} makes sure that salt-minion can always be updated safely.
+{suse} makes sure that [package]``salt-minion`` can always be updated safely.
 
 
 
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -107,3 +107,4 @@ the update of the main server (master server) is completed.
 
 Proxies should be updated after all server updates are complete.
 Especially in a version 3.2 to 4.0 migration you should upgrade the server first, then any proxy.
+For more information, see xref:upgrade:proxy-update.adoc[] and xref:upgrade:proxy-migration.adoc[].
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -1,33 +1,39 @@
 [[maintenance-window]]
-= Maintenance Window Checklist
+= {productname} Maintenance Window
 
-If you work on a {productname} Server with scheduled maintenance windows, you might find it difficult to remember all the things that you need to during that critical downtime.
+If you work with scheduled maintenance windows, you might find it difficult to remember all the things that you need to do before, during, and after that critical downtime of the {productname} Server.
+{productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have do be considered.
 
-As with any {sle} system, you are encouraged to install package updates on the {susemgr} Server when they arrive.
+{suse} recommends to keep the {productname} infrastructure (servers, proxies, build hosts, etc.) always up-to-date.
+Because the {productname} Server is needed to update all your IT landscape, the server must always be updated.
+// It's similar to zypper at the package level:
+You must keep it up-to-date or you will risk not being able to update something else successfully!
+
+
+
+== {productname} Server
 
 This section contains a checklist for your maintenance window, with links to further information on performing each of the steps, if you need it.
 
-
-. Stop services.
-You will need to stop the spacewalk, SAP, and database services, along with any others you have running.
+// ke, 2019-09-30: we'll stop spacewalk during the update
+// . Stop spacewalk services.
+// You will need to stop the spacewalk, SAP, and database services, along with any others you have running.
 . Check if the configuration is still correct.
-. On Salt systems, apply the highstate.
 . Apply the latest updates.
 See xref:upgrade:server-update.adoc[].
 . Update to the latest service pack, if required.
 See xref:upgrade:migrate-4x-4x.adoc[].
 . Reboot the server.
 . Check if the configuration is still correct.
-. On Salt systems, apply the highstate.
-. Start any stopped services.
+// . Start any stopped services.
 
+For information about database schema upgrades and PostgreSQL migrations, see xref:upgrades:db-migration.adoc[].
 
 You can install updates using your package manager.
 For information on using {yast}, see https://www.suse.com/documentation/sles-15/book_sle_admin/data/cha_onlineupdate_you.html).
 For information on using zypper, see https://www.suse.com/documentation/sles-15/book_sle_admin/data/sec_zypper.html).
 
 ////
-
 Preferable, you will run such a tool within a maintenance window; for more information, see xref:administration:maintenance-window.adoc#maintenance-window[].
 
  complete procedure, also see above:
@@ -44,7 +50,7 @@ Preferable, you will run such a tool within a maintenance window; for more infor
 By default, several update channels are configured and enabled for the {susemgr} Server.
 New and updated packages will become available automatically.
 
-To keep {susemgr} up to date, you must have it connected directly to {scc}.
+To keep {susemgr} up to date, you must have it connected directly to {scc}.
 Alternatively, for disconnected environments, you can use {rmtool} (RMT) as a local installation source.
 
 You can check that the update channels are available on your system with this command:
@@ -86,3 +92,16 @@ zypper migration
 ----
 
 This procedure is also known as a Service Pack (SP) migration.
+
+
+
+== Inter-Server Synchronization Slave Server
+
+Inter-server synchronization slave servers should be updated as soon as
+the update of the main server (master server) is completed.
+
+
+
+== {productname} Proxy
+
+Proxies should be updated after all Server updates are complete.
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -2,7 +2,7 @@
 = {productname} Maintenance Window
 
 If you work with scheduled maintenance windows, you might find it difficult to remember all the things that you need to do before, during, and after that critical downtime of the {productname} Server.
-{productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have do be considered.
+{productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have to be considered.
 
 {suse} recommends to keep the {productname} infrastructure (servers, proxies, build hosts, etc.) always up-to-date.
 Because the {productname} Server is needed to update all your IT landscape, the server must always be updated.
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -4,7 +4,7 @@
 If you work with scheduled maintenance windows, you might find it difficult to remember all the things that you need to do before, during, and after that critical downtime of the {productname} Server.
 {productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have to be considered.
 
-{suse} recommends to keep the {productname} infrastructure (servers, proxies, build hosts, etc.) always up-to-date.
+{suse} recommends to keep the {productname} infrastructure (Servers, Proxies, build hosts, etc.) always up-to-date.
 Because the {productname} Server is needed to update all your IT landscape, the server must always be updated.
 // It's similar to zypper at the package level:
 You must keep it up-to-date or you will risk not being able to update something else successfully!
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -51,8 +51,8 @@ Preferable, you will run such a tool within a maintenance window; for more infor
 By default, several update channels are configured and enabled for the {susemgr} Server.
 New and updated packages will become available automatically.
 
-To keep {susemgr} up to date, you must have it connected directly to {scc}.
-Alternatively, for disconnected environments, you can use {rmtool} (RMT) as a local installation source.
+To keep {susemgr} up to date, either connect it directly to {scc} or use {rmtool} (RMT).
+You can use RMT as a local installation source for disconnected environments.
 
 You can check that the update channels are available on your system with this command:
 
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -26,6 +26,7 @@ See xref:upgrade:migrate-4x-4x.adoc[].
 // . Reboot the server.
 // . Check if the configuration is still correct.
 // . Start any stopped services.
+. Run [command]``spacewalk-service status`` and check whether all required services are up and running.
 
 For information about database schema upgrades and PostgreSQL migrations, see xref:upgrades:db-migration.adoc[].
 
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -23,7 +23,8 @@ This section contains a checklist for your maintenance window, with links to fur
 See xref:upgrade:server-update.adoc[].
 . Update to the latest service pack, if required.
 See xref:upgrade:migrate-4x-4x.adoc[].
-. Reboot the server.
+// We reboot during the above listed procedures.
+// . Reboot the server.
 // . Check if the configuration is still correct.
 // . Start any stopped services.
 
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -86,13 +86,13 @@ For example, the major release 4.0 will be incremented to 4.0.1 when an MU is re
 You can verify which version you are running by looking at the bottom of the navigation bar in the {webui}.
 You can also fetch the version number with the [literal]``api.getVersion()`` XMLRPC API call.
 
-To upgrade to the latest MU, you can use the [command]``zypper migration`` command at the command prompt:
-
-----
-zypper migration
-----
-
-This procedure is also known as a Service Pack (SP) migration.
+// To upgrade to the latest MU, you can use the [command]``zypper migration`` command at the command prompt:
+// 
+// ----
+// zypper migration
+// ----
+// 
+// This procedure is also known as a Service Pack (SP) migration.
 
 
 
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -5,9 +5,8 @@ If you work with scheduled maintenance windows, you might find it difficult to r
 {productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have to be considered.
 
 {suse} recommends to keep the {productname} infrastructure (Servers, Proxies, build hosts, etc.) always up-to-date.
-Because the {productname} Server is needed to update all your IT landscape, the server must always be updated.
 // It's similar to zypper at the package level:
-You must keep it up-to-date or you will risk not being able to update something else successfully!
+You must keep the {productname} Server up-to-date or you will risk not being able to update some parts of your IT landscape successfully!
 
 
 
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -116,6 +116,7 @@ the update of the main server (master server) is completed.
 
 == {productname} Proxy
 
-Proxies should be updated after all server updates are complete.
+Proxies should be updated as soon as Server updates are complete.
+In general, running a Proxy connected to a Server on a different version is not supported. The only exception is for the duration of updates where it is expected that the Server is updated first, so the Proxy could run the previous version temporarily.
 Especially in a version 3.2 to 4.0 migration you should upgrade the server first, then any proxy.
 For more information, see xref:upgrade:proxy-update.adoc[] and xref:upgrade:proxy-migration.adoc[].
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -18,13 +18,13 @@ This section contains a checklist for your maintenance window, with links to fur
 // ke, 2019-09-30: we'll stop spacewalk during the update
 // . Stop spacewalk services.
 // You will need to stop the spacewalk, SAP, and database services, along with any others you have running.
-. Check if the configuration is still correct.
+// . Check if the configuration is still correct.
 . Apply the latest updates.
 See xref:upgrade:server-update.adoc[].
 . Update to the latest service pack, if required.
 See xref:upgrade:migrate-4x-4x.adoc[].
 . Reboot the server.
-. Check if the configuration is still correct.
+// . Check if the configuration is still correct.
 // . Start any stopped services.
 
 For information about database schema upgrades and PostgreSQL migrations, see xref:upgrades:db-migration.adoc[].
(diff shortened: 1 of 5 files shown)


modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
...
+== {productname} Server
+
+This section contains a checklist for your maintenance window, with links to further information on performing each of the steps, if you need it.
 
-. Stop services.
-You will need to stop the spacewalk, SAP, and database services, along with any others you have running.
+// ke, 2019-09-30: we'll stop spacewalk during the update
+// . Stop spacewalk services.
+// You will need to stop the spacewalk, SAP, and database services, along with any others you have running.
 . Check if the configuration is still correct.
```
juliogonzalez: I wonder what this means.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
...
-. On Salt systems, apply the highstate.
 . Apply the latest updates.
 See xref:upgrade:server-update.adoc[].
 . Update to the latest service pack, if required.
 See xref:upgrade:migrate-4x-4x.adoc[].
```
juliogonzalez: This link points to a document with the following introduction:
 > Migrating {productname} from version 4 to 4.x (when available) can be done using either the {yast} Online Migration tool, or with the Zypper command line tool.

We should maybe mention here:

> Update {productname} Server to the latest major version available, if required
> See xref:upgrade:migrate-4x-4x.adoc[].

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
...
 . Reboot the server.
```
juliogonzalez: This should be part of one of the two points above, FMPOV.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -44,7 +50,7 @@ Preferable, you will run such a tool within a maintenance window; for more infor
 By default, several update channels are configured and enabled for the {susemgr} Server.
 New and updated packages will become available automatically.
 
-To keep {susemgr} up to date, you must have it connected directly to {scc}.
+To keep {susemgr} up to date, you must have it connected directly to {scc}.
```
juliogonzalez: Well, you it's not a "must", as below we mention that you can also use RMT.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -86,3 +92,16 @@ zypper migration
...
+== Inter-Server Synchronization Slave Server
+
+Inter-server synchronization slave servers should be updated as soon as
+the update of the main server (master server) is completed.
+
+
+
+== {productname} Proxy
+
+Proxies should be updated after all Server updates are complete.
```
juliogonzalez: > Proxies should be updated after all {productname} Server updates are complete.

?

modules/upgrade/pages/migrate-4x-4x.adoc
```diff
@@ -7,8 +7,8 @@ Migrating {productname} from version 4 to 4.x (when available) can be done usin
 
 [IMPORTANT]
 ====
-Upgrades should be run from a text console, rather than a graphical interface like GNOME.
-If you are logged into a GNOME session running on the machine you are going to update, you will need to switch to a text console.
+Migrations should be run from a text console, rather than a graphical interface like GNOME.
```
juliogonzalez: Now that 4.x is a base product, we don't install WebUI by default, and I wonder if we should even support it.

modules/upgrade/pages/migrate-4x-4x.adoc
```diff
@@ -38,7 +38,7 @@ If the process fails, check these issues first:
 * If {yast} does not have the online migration module available, install the [package]``yast2-migration`` package.
 After installing, restart {yast} to ensure the module is available.
 * If there are older updates available, {yast} will notify you and ask to install them first.
-We strongly recommend you install all updates before performing the upgrade.
+We strongly recommend you install all updates before performing the migration.
```
juliogonzalez: > You must install all updates before performing the migration.

modules/upgrade/pages/migrate-4x-4x.adoc
```diff
@@ -7,8 +7,8 @@ Migrating {productname} from version 4 to 4.x (when available) can be done usin
```
juliogonzalez: I am quite confused about this document. Is this for migration the base OS or the product from 4 to 4.x?

I can see we don't talk, for example, about DB update.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
 [[maintenance-window]]
-= Maintenance Window Checklist
+= {productname} Maintenance Window
 
-If you work on a {productname} Server with scheduled maintenance windows, you might find it difficult to remember all the things that you need to during that critical downtime.
+If you work with scheduled maintenance windows, you might find it difficult to remember all the things that you need to do before, during, and after that critical downtime of the {productname} Server.
+{productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have do be considered.
```
moio:
```suggestion
{productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have to be considered.
```

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
...
+
+{suse} recommends to keep the {productname} infrastructure (servers, proxies, build hosts, etc.) always up-to-date.
```
moio:
```suggestion
{suse} recommends to keep the {productname} infrastructure (Servers, Proxies, build hosts, etc.) always up-to-date.
```

(I think we capitalize those, right @Loquacity?)

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
...
+Because the {productname} Server is needed to update all your IT landscape, the server must always be updated.
```
moio: Isn't this a repetition of the previous line?

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
...
+// It's similar to zypper at the package level:
+You must keep it up-to-date or you will risk not being able to update something else successfully!
 
-As with any {sle} system, you are encouraged to install package updates on the {susemgr} Server when they arrive.
 
-This section contains a checklist for your maintenance window, with links to further information on performing each of the steps, if you need it.
 
+== {productname} Server
+
+This section contains a checklist for your maintenance window, with links to further information on performing each of the steps, if you need it.
```
moio:
```suggestion
This section contains a checklist for your maintenance window, with links to further information on performing each of the steps.
```

modules/administration/pages/maintenance-window.adoc
moio: I guess both kinds of migrations should actually be mentioned: SP upgrades of the underlying SLES/openSUSE minor version and minor product upgrades.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,33 +1,39 @@
...
-. On Salt systems, apply the highstate.
 . Apply the latest updates.
 See xref:upgrade:server-update.adoc[].
 . Update to the latest service pack, if required.
 See xref:upgrade:migrate-4x-4x.adoc[].
 . Reboot the server.
 . Check if the configuration is still correct.
-. On Salt systems, apply the highstate.
-. Start any stopped services.
+// . Start any stopped services.
```
moio: I am confused why this is commented now.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -86,3 +92,16 @@ zypper migration
 ----
 
 This procedure is also known as a Service Pack (SP) migration.
+
+
+
+== Inter-Server Synchronization Slave Server
+
+Inter-server synchronization slave servers should be updated as soon as
```
moio:
```suggestion
Inter-server synchronization Slave Servers should be updated as soon as
```

modules/administration/pages/maintenance-window.adoc
```diff
@@ -86,3 +92,16 @@ zypper migration
...
+the update of the main server (master server) is completed.
```
moio:
```suggestion
the update of the main server (Master Server) is completed.
```

modules/administration/pages/maintenance-window.adoc
moio: I think this is correct, although it could be made clearer perhaps.

Basically we are saying that in a 3.2 to 4.0 migration you should upgrade the Server first, then any Proxy.

modules/upgrade/pages/migrate-4x-4x.adoc
moio: You mean a GUI?

modules/administration/pages/maintenance-window.adoc
juliogonzalez: FMPOV this should be removed, and made part of the update/migration procedures.

Here we could mention, maybe, that the user should use `spacewalk-service status` and check if all the services required are up and running (problem: not all users require the same services).

modules/administration/pages/maintenance-window.adoc
keichwa: me too, I'm going to delete it.

modules/administration/pages/maintenance-window.adoc
keichwa: I see your point and I'm mostly for it.  But because SUSE Manger now is are product (and not just an Extension), I'm wondering whether we still want to talk that much about the underlying OS.  It probably makes sense if "zypper migration" will still offer such migration when available:
```
SLES 15 SP1 with SUMA 4.0
  |
  +-> SLES 15 SP2 with SUMA 4.0
```

modules/administration/pages/maintenance-window.adoc
keichwa: yes

modules/upgrade/pages/migrate-4x-4x.adoc
keichwa: this is about maintenace window as a concept (buzzword) ... schedule downtime, do the "real things", become available again.

The "real things" are discussed in the linked files such as upgrade:server-update.adoc and upgrade:migrate-4x-4x.adoc.  Of course, everything is connected somehow, but I do not want to discuss everything everywhere ;)

modules/administration/pages/maintenance-window.adoc
keichwa: yes, to be cap'ed if they appear close with {productname}

modules/upgrade/pages/migrate-4x-4x.adoc
juliogonzalez: No, what I mean is that I don't see anything about the DB update at ` upgrade:migrate-4x-4x.adoc` :-)

keichwa: ah, ok.  will take a closer look and probably reference server-update.adoc.

modules/administration/pages/maintenance-window.adoc
keichwa: changed it

modules/upgrade/pages/migrate-4x-4x.adoc
keichwa: I think we cannot prevent users from installing a graphical desktop plus the graphical yast on the server.  As part of the installation documentation we actually mention that the user should select appropriate packages to run the graphical yast.  See https://github.com/SUSE/doc-susemanager/blob/master/modules/installation/pages/install-server-unified.adoc, step 13.

If this is not wanted, we should remove it everywhere.

modules/upgrade/pages/migrate-4x-4x.adoc
keichwa: done.

modules/upgrade/pages/migrate-4x-4x.adoc
keichwa: > No, what I mean is that I don't see anything about the DB update at  `upgrade:migrate-4x-4x.adoc` :-)

Because you are asking I assume neither `yast migration` nor `zypper migration` run the schema upgrade?  Please confirm :) @juliogonzalez

modules/administration/pages/maintenance-window.adoc
keichwa: ok, done

modules/administration/pages/maintenance-window.adoc
keichwa: @juliogonzalez 
> We should maybe mention here:
> 
> > Update {productname} Server to the latest major version available, if required
> > See xref:upgrade:migrate-4x-4x.adoc[].

Hhmm, no or I'm not sure what you mean :) The upgrade scenario ("migration") usually does not work or is not recommended for major version upgrades -- for this, you need the offline procedure with autoyast (or simply booting from a boot medium and then run the update/upgrade). This is what we have in migrate-3x-4x.adoc (IMO).

If things become available we should describe them as needed.  SUMA 5 on SLES 15.x, maybe. SLES 16 running SUMA 4.x, probably not, but who knows...

The main problem usually is that nobody informs us early enough, and nobody paints the complete picture.  Of course, nobody knows the complete picture. And those who know a lot of it, often cannot imagine how stupid we are ;)

modules/administration/pages/maintenance-window.adoc
keichwa: @Loquacity this is a corner case.  Please help!

modules/administration/pages/maintenance-window.adoc
keichwa: @Loquacity this too

modules/administration/pages/maintenance-window.adoc
keichwa: i'm now mentioning this.  let's see whether people will like it.

keichwa:
All my edits are in.  There are more open questions above.

modules/administration/pages/maintenance-window.adoc
juliogonzalez: Yeah, "latest major version available" is not the correct sentence. It should be "latest Maintenance Update available for your current major version`.

Let me explain it more clearly.

Talking about SUSE Manager, we basically use a `X.Y.Z` versioning.

`Z` changes: that means a MU. User stops services, updates packages, runs the script to update the database, starts services, end of the story: Example 4.0.1 -> 4.0.2

`Y` changes: Means that a new major version, but that allows an in-line "migration". You don't replace the server. It usually means you will jump to a new SLE service pack. Bue before starting you need a fully updated available `Z` version of your `Y`. Example: 3.1.11 to 3.2.0.

`X` changes: Means that's a breaking change, that doesn't allow an in-line "migration". You need your current server updated to the last `Y.Z` of your `X` version. And you need a new server with the new version, so you can migrate your data. Example: 3.2.11 to 4.0.0.

So FMPOV that `xref:upgrade:migrate-4x-4x.adoc[]` is like a document for the future, describing what happens when `Y` changes. That is: when we will have `4.1.0` available, so users can migrate from `4.0.X`, as it says:

> Migrating {productname} from version 4 to 4.x (when available) can be done using

And where you say `4' , I guess it should say `4.0`.

In this case, the users will need to be at the latest MU available for `4.0` before being able to migrate to `4.1`

Right now we need to have guides for "Z changes" (MUs), "X changes" (migration from 3.2 to 4.0), and you can already prepare "Y changes" (4.0 -> 4.1), but I think we should not offer it to clients yet.

keichwa:
I just discovered that that after the list of items we have interesting info that I'd like to move to the two other files.  If you veto please raise your voice!

modules/upgrade/pages/migrate-4x-4x.adoc
juliogonzalez: That's right, we can't prevent users from installing a GUI and graphical yast2.

But if we mention it, then users and on their own to find how to get some things, because we don't explain how to do it with console yast2 and GUI yast.

Example: https://github.com/SUSE/doc-susemanager/blob/master/modules/installation/pages/server-setup.adoc

> Log in to the {productname} Server and type yast2 susemanager_setup to begin the setup.
> From the introduction screen select menu:SUSE Manager Setup[Setup SUSE Manager from scratch] and click btn:[Next] to continue.

@moio: what's your take here?

modules/administration/pages/maintenance-window.adoc
keichwa: "Z changes" (MUs) are accumulated and consolidated package updates, IMO.  And this is what I want to see covered in the server-update.adoc file.  I basically agree with all you said here :)

modules/administration/pages/maintenance-window.adoc
Loquacity: I don't see any reason to initcap here. It's not a proper noun.

Loquacity: > I don't see any reason to initcap here. It's not a proper noun.

modules/administration/pages/maintenance-window.adoc
Loquacity: no initcaps.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -79,10 +86,36 @@ For example, the major release 4.0 will be incremented to 4.0.1 when an MU is re
...
+
+
+
+== {productname} Proxy
+
+Proxies should be updated after all server updates are complete.
```
moio:
```suggestion
Proxies should be updated as soon as Server updates are complete.
In general, running a Proxy connected to a Server on a different version is not supported. The only exception is for the duration of updates where it is expected that the Server is updated first, so the Proxy could run the previous version temporarily.
```

modules/administration/pages/maintenance-window.adoc
moio: Server is short of `{productname} Server`, don't we initcap that?

Also, ISS Masters and Slaves are initcapped everywhere in the UI:

Screenshot from 2019-10-03 09-32-06

and API doc:

Screenshot from 2019-10-03 09-32-52


...so I thought we should stay consistent.

modules/administration/pages/maintenance-window.adoc
moio: Yes @keichwa.

There's two extra twists to keep in mind: underlying OS SP migrations and Postgres migrations are typically done in between "Y changes".

An example progression from the past versions could be:

 - 3.1.4, SLES 12 SP3, Postgres 9.6
 - 3.1.11, SLES 12 SP3, Postgres 9.6
 - 3.2.10, SLES 12 SP3, Postgres 9.6
 - 3.2.10, SLES 12 SP3, Postgres 10
 - 3.2.10, SLES 12 SP4, Postgres 10
 - 4.0.1, SLES 15 SP1, Postgres 10

Package upates only cover "Z changes" (MUs), SP migrations and "Y changes" need `zypper migrate`, Postgres needs the old and new versions to be installed in parallel plus a manual procedure to migrate data, and "X changes" require a new server altogether.

Hopefully all cases are covered above!

modules/upgrade/pages/migrate-4x-4x.adoc
moio: Schema upgrades are indeed a separate step not triggered by any package, so `zypper` will not perform them.

modules/upgrade/pages/migrate-4x-4x.adoc
moio: Uyuni users can do whatever they prefer, but as far as SUSE Manager is concerned it is supported as an appliance only, meaning no extra packages outside of a handful exceptions.

Running a GUI on top of the SUSE Manager Server is not something we support at all AFAIK.

keichwa:
On 2019-10-03 09:54, Silvio Moioli wrote:

> Running a GUI on top of the SUSE Manager Server is not something we
> support at all AFAIK.

Makes sense, but it was different in the past.  Needs further 
clarification (will cause more edits elsewhere in our docs).

-- 
Karl Eichwalder

modules/upgrade/pages/migrate-4x-4x.adoc
keichwa: Joe is fine with running a minimal desktop on the server.  Thus closing this for now.

modules/administration/pages/maintenance-window.adoc
keichwa: ok, this again is a @Loquacity remark.  Of course, I vote to be consistent as possible.

modules/administration/pages/maintenance-window.adoc
keichwa: I added a remark about the database schema upgrade.

Ths version info went into the db-migration file. For the moment, it's commented.

keichwa:
I'd like to merge it soon. Otherwise a fear that conflicts will arise. Can I have reviews, please?

When merged, we can start a new PR to address the rest.

modules/administration/pages/maintenance-window.adoc
Loquacity: I would argue that the UI is wrong. Basically, initcaps make text harder to read, so we should avoid them wherever possible. Also, the argument that "server" is short for "{productname} Server" is silly, because grammar doesn't work like that. "Server" on its own is a noun, and adding the {productname} in makes it a proper noun. SUSE does not own the word "Server", they own the noun phrase "SUSE Manager Server".

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,36 +1,44 @@
 [[maintenance-window]]
-= Maintenance Window Checklist
+= {productname} Maintenance Window
```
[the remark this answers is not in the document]
keichwa: ok, removing {productname} now

modules/upgrade/pages/migrate-4x-4x.adoc
keichwa: ok, I added some such remark.  I hope it is clear enough now.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,36 +1,45 @@
...
-= Maintenance Window Checklist
+= Maintenance Window
 
-If you work on a {productname} Server with scheduled maintenance windows, you might find it difficult to remember all the things that you need to during that critical downtime.
+If you work with scheduled maintenance windows, you might find it difficult to remember all the things that you need to do before, during, and after that critical downtime of the {productname} Server.
+{productname} Server related systems such as Inter-Server Synchronization Slave Servers or {productname} Proxies are also affected and have to be considered.
 
-As with any {sle} system, you are encouraged to install package updates on the {susemgr} Server when they arrive.
+{suse} recommends you always keep your {productname} infrastructure updated.
+That includes servers, proxies, and build hosts.
```
juliogonzalez: Thinking about this: if we mention build host, I wonder if we shouldn't mention monitoring host (the clients with grafana and prometheus). @cavalheiro @renner?

modules/administration/pages/maintenance-window.adoc
```diff
@@ -1,36 +1,45 @@
...
+// It's similar to zypper at the package level:
+If you do not keep the {productname} Server updated, you might not be able to update some parts of your environment when you need to.
```
juliogonzalez: Only the server? Maybe we should say better `{productname} infrastructure`?

modules/administration/pages/maintenance-window.adoc
```diff
@@ -44,8 +53,8 @@ Preferable, you will run such a tool within a maintenance window; for more infor
 By default, several update channels are configured and enabled for the {susemgr} Server.
 New and updated packages will become available automatically.
 
-To keep {susemgr} up to date, you must have it connected directly to {scc}.
-Alternatively, for disconnected environments, you can use {rmtool} (RMT) as a local installation source.
+To keep {susemgr} up to date, either connect it directly to {scc} or use {rmtool} (RMT).
```
juliogonzalez: We don't want this for Uyuni, so this should be inside an `if content_suma`.

modules/upgrade/pages/db-migration.adoc
```diff
@@ -12,6 +12,21 @@ You cannot migrate directly from PostgreSQL{nbsp}9.4 to version 10.
...
+////
+== Background Information
+
+
+* {productname} 3.1.4, SLES 12 SP3, Postgres 9.6
+* {productname} 3.1.11, SLES 12 SP3, Postgres 9.6
+* {productname} 3.2.10, SLES 12 SP3, Postgres 9.6
+* {productname} 3.2.10, SLES 12 SP3, Postgres 10
+* {productname} 3.2.10, SLES 12 SP4, Postgres 10
+* {productname} 4.0.1, SLES 15 SP1, Postgres 10
```
juliogonzalez: Better change thos to 4.0.0, SLES 15 SP1, Postgres 10 :-)

modules/upgrade/pages/migrate-4x-4x.adoc
```diff
@@ -1,32 +1,46 @@
 [[bp.sp.migration.version]]
-= Migrating from Version 4 to 4.x
+= Upgrading from Version 4 to 4.x
 
-Migrating {productname} from version 4 to 4.x (when available) can be done using either the {yast} Online Migration tool, or with the Zypper command line tool.
+You can upgrade {productname} from version 4 to 4.x (when available) using either the {yast} online migration tool or the Zypper command line tool.
```
juliogonzalez: Why are we renaming this from `migration` to `update` if in the end we use `yast2-migration`?

modules/upgrade/pages/migrate-4x-4x.adoc
```diff
@@ -74,6 +93,13 @@ spacewalk-service start
 
 If the process fails, check these issues first:
 
-* If Zypper  does not have the migration tool available, install the [package]``zypper-migration-plugin`` package.
+* If Zypper does not have the migration tool available, install the [package]``zypper-migration-plugin`` package.
```
juliogonzalez: Same here, we are using `zypper-migration-plugin`, so renaming the procedures to `upgrade` seems to be confusing.

modules/upgrade/pages/db-migration.adoc
keichwa: ok, but it is currently commented...

modules/upgrade/pages/migrate-4x-4x.adoc
keichwa: Yes, this is not nice.  But we said we want to call the small updates (z level) such as package updates or MUs _updates_, the y level updates (4.0.z -> 4.1.0) _upgrades_, and the x level updates (3.y.z -> 4.0.0) _migrations_.

Unfortunately, the y level updates are also known as "SP migration"...  @Loquacity do we have to improve our terminology?

modules/administration/pages/maintenance-window.adoc
keichwa: moio argued that it is especially important to keep the Server updated, the rest can follow.

juliogonzalez: FMPOV: Server and Proxy, the rest can follow.

But if we mention build hosts, we should mention monitoring hosts as well (or whatever the name we have for the hosts where we install grafana and the prometheus server)

modules/administration/pages/maintenance-window.adoc
cavalheiro: As of now, we still do not have provisioning of Prometheus and Grafana servers. But its worth mentioning once we have it.

juliogonzalez: Not automated, but users can still install grafana and prometheus on clients following manual instructions from the doc, right?

If so, they should update such clients with grafana and prometheus together with the server, the proxy and the build hosts.

cavalheiro: That's right, I agree.

modules/upgrade/pages/migrate-4x-4x.adoc
Loquacity: Oh, probably. But there's no consensus anywhere on this. If the system or the UI was consistent, we could use those terms, but they aren't. So my vote is to standardise within our docs and hope that everything else catches up.

keichwa: Never give up :)  We must pay attention when they start changing something in that area and and propose our ideas as early as possible.

PR #789, opened 2019-10-04
make systemdata copying an optionally substep of the migration

keichwa:
Hubert, do you think it makes sense to restructure this chapter this way?  I'd like to have the 'mgr-setup -r' step part of the "normal" migration procedure, just optionally.

### DIFF
--- a/modules/upgrade/pages/migrate-3x-4x.adoc
+++ b/modules/upgrade/pages/migrate-3x-4x.adoc
@@ -22,6 +22,7 @@ We strongly recommend that you contact SUSE Consulting to assist with this proce
 ====
 
 
+
 == Prepare to Migrate
 
 The source system must be running {productname} 3.2 with all the latest updates applied.
@@ -33,7 +34,7 @@ It is important that PostgreSQL 10 is already running on your {productname} 3.
 For more information, see xref:db-migration.adoc[].
 ====
 
-During migration, the database on the source system needs to get exported, or dumped.
+During migration, the database on the source system needs to get dumped.
 The dump is compressed, and temporarily stored on the target system.
 The compression is done using [command]``gzip`` using the default compression options.
 Maximum compression only yields about 10% of space savings.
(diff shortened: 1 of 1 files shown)


keichwa:
@thallma , unfortunately I cannot add you here as a reviewer...

modules/upgrade/pages/migrate-3x-4x.adoc
```diff
@@ -33,7 +34,7 @@ It is important that PostgreSQL 10 is already running on your {productname} 3.
 For more information, see xref:db-migration.adoc[].
 ====
 
-During migration, the database on the source system needs to get exported, or dumped.
+During migration, the database on the source system needs to get dumped.
```
mantel: I like the term "exported" better than "dumped". A database dump typically is a binary dump, but so it can only be imported into a database with the exactly same version. An export contains the whole database as SQL statements, so importing is much slower, but it works across different database versions. Not really important, but technically experienced users might expect something else when reading "dump".

keichwa: That's a perfectly valid concern!

modules/upgrade/pages/migrate-3x-4x.adoc
```diff
@@ -108,20 +112,27 @@ When you have received the `migration successful`` message, you need to reconfig
...
 
-== Copy Data to the Target System
+
+[[migration.troubleshooting]]
+== Troubleshooting
+
+
+
+[[migration.troubleshooting.systemdata]]
+==== Copy System Data to the Target System
```
Loquacity: Double heading :(

modules/upgrade/pages/migrate-3x-4x.adoc
```diff
@@ -144,6 +155,10 @@ While the data migration is in progress, the database services will be shut down
 This is to ensure that no data is written to the database during the migration.
 
 
+
+[[migration.troubleshooting.pkgdata]]
+=== Intergrate Externally Stored Package Data
```
Loquacity:
```suggestion
=== Integrate Externally Stored Package Data
```

modules/upgrade/pages/migrate-3x-4x.adoc
keichwa: workaround is in

keichwa:
@Loquacity please better check my English phrases again ;)

PR #790, opened 2019-10-07
Technical review: Admin guide: Inter Server Sync

hustodemon:
fixes for minor errors encountered during the review.

TODO after merging: check that the FQDNs (`server1.example.com`) are **not** rendered as links. This was the case before and it was wrong:

ss


Loquacity:
@hustodemon do you need one of us to merge?

PR #791, opened 2019-10-07
Xref ia repair

jcayouette:


Loquacity:
Can you please make sure you wait for reviews before merging? It would be nice to have everyone following the same process, if we can.

jcayouette:
Sure, I certainly agree with you but since these were simple changes I used my superpowers so we could get our CI online.

PR #804, opened 2019-10-09
maint window and upgrade

keichwa:
per feedback, see
https://github.com/SUSE/doc-susemanager/pull/781#discussion_r332868061

### DIFF
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -53,8 +53,10 @@ complete procedure, also see above:
 By default, several update channels are configured and enabled for the {susemgr} Server.
 New and updated packages will become available automatically.
 
+ifeval::[{suma-content} == true]
 To keep {susemgr} up to date, either connect it directly to {scc} or use {rmtool} (RMT).
 You can use RMT as a local installation source for disconnected environments.
+endif::[]
 
 You can check that the update channels are available on your system with this command:
 
--- a/modules/administration/pages/maintenance-window.adoc
+++ b/modules/administration/pages/maintenance-window.adoc
@@ -50,7 +50,7 @@ complete procedure, also see above:
     spacewalk-service start
 ////
 
-By default, several update channels are configured and enabled for the {susemgr} Server.
+By default, several update channels are configured and enabled for the {productname} Server.
 New and updated packages will become available automatically.
 
 ifeval::[{suma-content} == true]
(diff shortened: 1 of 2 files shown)


modules/administration/pages/maintenance-window.adoc
```diff
@@ -53,8 +53,10 @@ complete procedure, also see above:
 By default, several update channels are configured and enabled for the {susemgr} Server.
 New and updated packages will become available automatically.
 
+ifeval::[{suma-content} == true]
 To keep {susemgr} up to date, either connect it directly to {scc} or use {rmtool} (RMT).
 You can use RMT as a local installation source for disconnected environments.
+endif::[]
 
 You can check that the update channels are available on your system with this command:
```
juliogonzalez: Unrelated to this PR?

The content below only applies to SUSE Manager, not Uyuni. As we neither have MUs, I'd say you could move the `endif::[]` so it applies until line 99.

modules/administration/pages/maintenance-window.adoc
```diff
@@ -53,8 +53,10 @@ complete procedure, also see above:
 By default, several update channels are configured and enabled for the {susemgr} Server.
```
juliogonzalez: Un related to this PR, shouldn't we use `{productname}` above?

modules/administration/pages/maintenance-window.adoc
keichwa: ok, but ifdefs must not exceed more than 1 page (~20 lines) ;)  That's my very personal rule. I either will create an include file or create 1 file file for suma and the other one for uyuni (and put the ifdef in the upper-level nav-file).

modules/administration/pages/maintenance-window.adoc
keichwa: probably yes.  We sometimes to drive-by edits ;)

modules/administration/pages/maintenance-window.adoc
keichwa: done

modules/administration/pages/maintenance-window.adoc
keichwa: done

modules/upgrade/pages/db-migration.adoc
```diff
@@ -4,10 +4,10 @@
 This is about PostgreSQL database migration.
 
 If you want to upgrade {productname} 3.2 to {productname} 4 you must check whether {productname} 3.2 is already based on PostgreSQL version 10.
```
juliogonzalez: Not for this PR: I'd rewrite this doc so it fits both SUSE Manager or Uyuni as:

```
If you want to upgrade to the latest {productname} version you must check whether {productname} Server is already based on PostgreSQL version 10.
```

And then below, remove all mentions to `3.2` except where it's REALLY needed (and in that case enclose that with a diff).

PR #805, opened 2019-10-09
Update server-update.adoc

mantel:
Fix start command for the postgresql database.


keichwa:
> Fix start command for the postgresql database.

yes, but sigh.  I once tried to settle everything on the `service` variant, such as:
```
service postgresql start
```
But maybe, we now better switch to `systemctl`.  I've no strong preference.

PR #808, opened 2019-10-09
Update migrate-3x-4x.adoc

mantel:
This probably should have been part of the previous commit. As explained there, the migration script will always implicitly run the rsync (that is triggered manually via "mgr-setup -r"), so there is no point running it again after the successful migration.


keichwa:
> This probably should have been part of the previous commit. As explained there, the migration script will always implicitly run the rsync (that is triggered manually via "mgr-setup -r"), so there is no point running it again after the successful migration.

I think we can first push it to `mantel-patch-4`.  Trying...

PR #810, opened 2019-10-10
Prometheus unsupported in non-routable networks

Loquacity:
https://github.com/SUSE/spacewalk/issues/9413

### DIFF
--- a/modules/administration/pages/monitoring.adoc
+++ b/modules/administration/pages/monitoring.adoc
@@ -7,7 +7,6 @@ You can monitor your {productname} environment using Prometheus and Grafana.
 [NOTE]
 ====
 Prometheus fetches metrics using a pull mechanism, so it must be able to establish network communications to monitored clients.
-If your clients are using a non-routable network with a reserved IP range, they will not be able to be contact the Prometheus server.
 ====
 
 
--- a/modules/administration/pages/monitoring.adoc
+++ b/modules/administration/pages/monitoring.adoc
@@ -6,7 +6,7 @@ You can monitor your {productname} environment using Prometheus and Grafana.
 
 [NOTE]
 ====
-Prometheus fetches metrics using a pull mechanism, so it must be able to establish network communications with monitored clients.
+Prometheus fetches metrics using a pull mechanism, so it must be able to establish network communications to monitored clients.
 If your clients are using a non-routable network with a reserved IP range, they will not be able to be contact the Prometheus server.
 ====
 
--- a/modules/administration/pages/monitoring.adoc
+++ b/modules/administration/pages/monitoring.adoc
@@ -6,8 +6,7 @@ You can monitor your {productname} environment using Prometheus and Grafana.
 
 [NOTE]
 ====
-{productname} does not currently support monitoring with Prometheus in private, non-routable networks.
-Prometheus must be able to establish network communications with monitored clients.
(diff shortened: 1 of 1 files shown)


modules/administration/pages/monitoring.adoc
```diff
@@ -4,6 +4,14 @@
 You can monitor your {productname} environment using Prometheus and Grafana.
 {productname} Server and Proxy are able to provide self-health metrics, or install and manage a limited number of Prometheus exporters on Salt clients.
 
+[NOTE]
+====
+{productname} does not currently support monitoring with Prometheus in private, non-routable networks.
```
cavalheiro: Remove this sentence

modules/administration/pages/monitoring.adoc
```diff
@@ -4,6 +4,14 @@
...
+Prometheus must be able to establish network communications with monitored clients.
```
cavalheiro: I would write: Since Prometheus fetches metrics based on a 'pull model', it needs to be able to open network connections to the monitored clients.

modules/administration/pages/monitoring.adoc
```diff
@@ -4,6 +4,13 @@
 You can monitor your {productname} environment using Prometheus and Grafana.
 {productname} Server and Proxy are able to provide self-health metrics, or install and manage a limited number of Prometheus exporters on Salt clients.
 
+[NOTE]
+====
+Prometheus fetches metrics using a pull mechanism, so it must be able to establish network communications with monitored clients.
```
cavalheiro: We should mention 'to' monitored clients instead of 'with'. The direction of the connection goes from the Prometheus server *to* the client

modules/administration/pages/monitoring.adoc
```diff
@@ -4,6 +4,13 @@
...
+If your clients are using a non-routable network with a reserved IP range, they will not be able to be contact the Prometheus server.
```
cavalheiro: This is not correct. The clients never contact the Prometheus server. It is the Prometheus server that contacts the clients, and that is were the problem is. Also, the reserved IP range is not meaningful here, and the definition of non-routable network can be confusing. I would remove this sentence altogether.

modules/administration/pages/monitoring.adoc
jcayouette: @cavalheiro Fixed.

modules/administration/pages/monitoring.adoc
jcayouette: @cavalheiro Removed the problem sentence.

Loquacity:
With these edits, you've actually removed the point of the admonition. The correct response to this as it stands is "so what?". I would argue that if this single sentence is the only part of the admonition we can agree on, it shouldn't be there at all.

cavalheiro:
We can rewrite the text to explain more in detail how Prometheus works and which network requirements it has. I would argue the original sentence 'If your clients are using a non-routable network with a reserved IP range, they will not be able to be contact the Prometheus server.' is not clear and not accurate, hence my suggestion to remove it.

jcayouette:
What about something like? 

> Prometheus fetches metrics using a pull mechanism, so the server must be able to establish network communications to monitored clients. Clients must have an open port and be reachable on the network.

cavalheiro:
> What about something like?
> 
> > Prometheus fetches metrics using a pull mechanism, so the server must be able to establish network communications to monitored clients. Clients must have an open port and be reachable on the network.

Sounds good to me!

PR #813, opened 2019-10-14
Update db-migration.adoc

mantel:


keichwa:
> I would not talk about postgresql 9.4 at all. It is so out of date, nobody should (and in fact can) still use it.

But we had such a case just 2 or 3 weeks ago! Someone wanted to upgrade from 3.1 to 4.0

Loquacity:
> > I would not talk about postgresql 9.4 at all. It is so out of date, nobody should (and in fact can) still use it.
> 
> But we had such a case just 2 or 3 weeks ago! Someone wanted to upgrade from 3.1 to 4.0

This is a very good point. Just because no one *should* be using it, doesn't mean no one is :wink:

PR #815, opened 2019-10-14
Registration: add a line for registering against the proxy

admd:
This page is also in good condition. It was just missing a line about the proxy field in case, the user wants to register the system against it.

### DIFF
--- a/modules/client-configuration/pages/registration-webui.adoc
+++ b/modules/client-configuration/pages/registration-webui.adoc
@@ -14,6 +14,7 @@ By default, the SSH  port is [systemitem]``22``.
 By default, the username is [systemitem]``root``.
 . In the [guimenu]``Password`` field, type password to log in to the client.
 . In the [guimenu]``Activation Key`` field, select the activation key that is associated with the software channel you want to use to bootstrap the client.
+. OPTIONAL: In the [guimenu]``Proxy`` field, select the proxy to register the client to.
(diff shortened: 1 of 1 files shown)


keichwa:
Thanks for noting, Abid!  We probably must indicate that this is "optional".  I'm sure @Loquacity will find good wording!

modules/client-configuration/pages/registration-webui.adoc
```diff
@@ -14,6 +14,7 @@ By default, the SSH  port is [systemitem]``22``.
 By default, the username is [systemitem]``root``.
 . In the [guimenu]``Password`` field, type password to log in to the client.
 . In the [guimenu]``Activation Key`` field, select the activation key that is associated with the software channel you want to use to bootstrap the client.
+. In the [guimenu]``Proxy`` field, select the proxy from the available ones to register the system against it.
```
Loquacity:
```suggestion
. OPTIONAL: In the [guimenu]``Proxy`` field, select which proxy to register the client against.
```

Loquacity: :question:

keichwa: yes, but ... I have the feeling that ,ost of the time we say "register to", "against" seems to occur 6 x.

Loquacity: You're right. I'll edit.

admd:
Thank you @Loquacity & @keichwa. I have changed the line according to the suggestions.

PR #818, opened 2019-10-14
move disconnected setup to administration from client-config

keichwa:
close https://github.com/SUSE/spacewalk/issues/9693


keichwa:
Wondering whether we should also add a redirect instead of https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/client-configuration/disconnected-setup.html?  And, if yes, how we want to manage such redirects?

Loquacity:
> Wondering whether we should also add a redirect instead of https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/client-configuration/disconnected-setup.html? And, if yes, how we want to manage such redirects?

As per discussion on the docs meeting, I don't think this is wise at this stage. Once we get the docs more stable, we can consider it.

keichwa:
I hope I properly merged it...  There was a strange conflict...

PR #826, opened 2019-10-16
Update db-migration.adoc

keichwa:
Thanks, but I think we'd better keep the info on 9.4 for the moment.


keichwa:
> Fine with me. IMHO it think it is a no-op now :)

yes, but you never know for sure that there still isn't a 3.1 system with postgresql 9.4 "running" somewhere. Approx. 2 months ago, I had to work on such an issue, and the user was complaining that we did not mention 9.4 in our documentation.  Of course, the user could have checked the 3.2 books... ;-)

PR #836, opened 2019-10-23
Swap space recommendation

admd:
With this PR, we provided a recommended value for swap space. I wasn't sure if some more explanation should be added here.

### Fixes
https://github.com/SUSE/spacewalk/issues/9821


modules/installation/pages/hardware-requirements.adoc
```diff
@@ -22,6 +22,7 @@ For {smr} hardware requirements, see xref:retail:retail-requirements.adoc[].
 | Disk Space:              | [path]``/`` _(root)_ Minimum 24{nbsp}GB
 |                          | [path]``/var/lib/pgsql`` Minimum 50{nbsp}GB
 |                          | [path]``/var/spacewalk`` Minimum 50{nbsp}GB per SUSE product and 360{nbsp}GB per Red Hat product
+| Swap space:               | 2-3{nbsp}GB
```
moio: How about just settling on a value, let's say 3 to stay abundant?

admd: Changed it to 3GB.

PR #838, opened 2019-10-23
Removal of broker section

brejoc:
I've opened this as a draft, so that we have a place to discuss this change. The problem I have with the broker description is the following: We don't really use the term broker anywhere and the description is confusing. P2P communication is not really something we leverage. Instead of we should talk about salt proxies and that proxies are used to scale large systems, because we can take load from the master to the proxies. Does that sound right you you, @Loquacity?


Loquacity:
That sounds perfectly reasonable to me. Would you like to draft something in there about proxies?

brejoc:
Awesome. Yes, I will try to come up with something today or tomorrow.

Loquacity:
@brejoc What's happening with this one?

PR #845, opened 2019-10-25
update screenshots

Loquacity:
https://github.com/SUSE/spacewalk/issues/9358


Loquacity:
image
image
image

Loquacity:
> Do we want to use the server name doc-test? That is fine with me but I suppose we need to be consistent with its use. We may also want to adjust the name of the ref client in the last image? You can do this using the dev tools in chrome. :) Otherwise LGTM!

There is zero consistency in any of reference guide. I think we need to address that as part of the epic instead.

jcayouette:
> > Do we want to use the server name doc-test? That is fine with me but I suppose we need to be consistent with its use. We may also want to adjust the name of the ref client in the last image? You can do this using the dev tools in chrome. :) Otherwise LGTM!
> 
> There is zero consistency in any of reference guide. I think we need to address that as part of the epic instead.

When I initially created our images with fennek my test setup was pretty consistent. But I suppose we need an update, agreed!

PR #853, opened 2019-10-29
SSL Certs - Replace on Existing

Loquacity:
https://github.com/SUSE/spacewalk/issues/9787

* Renames section from "Custom SSL" to "SSL Certificates", and re-writes intro to reflect contents more accurately.
* Adds new procedure from https://github.com/SUSE/spacewalk/files/3763488/pdfdoXPB59XUy.pdf


modules/administration/pages/ssl-certs.adoc
```diff
@@ -26,6 +30,8 @@ Ensure that the intermediate certificate comes first in the combined file.
 By default, {productname} uses a self-signed certificate.
 After you have completed the initial setup, you can replace the default certificate with a custom certificate.
 
+
```
keichwa: Because of back-ports (maybe) let's try avoid white-space changes as much as possible.  Esp. if we are not that sure whether we want to have 3 empty lines before procedure titles or not ;)

modules/administration/pages/ssl-certs.adoc
```diff
@@ -294,3 +305,59 @@ To complete the process, you need to update the database with this command:
...
+== Replace Existing CA and Server Certificates
+
+You can replace working certificates on an existing {productname} installation with a new third party certificate.
+In this case, you can replace the old CA certificate RPM with the new one, and update the database.
+
+
+
+.Procedure: Replacing Existing Certificates
+
+. On the {productname} Server, at the command prompt, move the old certificate directory to a new location:
```
keichwa: This is correct, but confusing (old - new - old) and in the end, you call the new directory old-something ;)
Suggestion (and remove the next lines):

```suggestion
. On the {productname} Server, at the command prompt, move the old certificate directory to a backup location:

----
mv /root/ssl-build /root/backup-ssl-build
----
```

I did not check whether `/root/old-ssl-build` occurs later on.

modules/administration/pages/ssl-certs.adoc
Loquacity: I agree, trying to add some consistency. I thought we'd agreed on three lines before headings?

keichwa:

> I agree, trying to add some consistency. I thought we'd agreed on three
> lines before headings?

Ok, but I thought it would apply for Chapter and Section titles only.
Good that I asked ;)  Also for admonitions and procedures?

-- 
                           90409 Nürnberg, Germany

Loquacity:
> I agree, trying to add some consistency. I thought we'd agreed on three lines before headings?
> Ok, but I thought it would apply for Chapter and Section titles only. Good that I asked ;) Also for admonitions and procedures?

Procedures, yes. Admonitions, I'm not so sure. They're not headings, they're block elements. We only give code blocks a single space.

modules/administration/pages/ssl-certs.adoc
```diff
@@ -1,9 +1,13 @@
-[[custom-ssl]]
-= Custom SSL Certificates
+[[ssl-certs]]
+= SSL Certificates
 
-You can use custom certificates with {productname} and {productname} Proxy.
+{productname} uses SSL certificates to ensure that clients are registered to the correct server.
+The host name of the SSL keys and certificates must match the fully qualified host name of the client system you deploy them on.
```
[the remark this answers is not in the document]
Loquacity: This is a really good explanation, I've added it to the concept at the top of the section.

modules/administration/pages/ssl-certs.adoc
```diff
@@ -294,3 +305,63 @@ To complete the process, you need to update the database with this command:
 ----
 
 If you use bootstrap, remember to also update your bootstrap scripts to reflect the new certificate information.
+
+
+
+== Replace Existing CA and Server Certificates
```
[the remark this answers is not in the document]
Loquacity: Yes, that was my assumption.

modules/administration/pages/ssl-certs.adoc
```diff
@@ -294,3 +305,63 @@ To complete the process, you need to update the database with this command:
...
+
+You can replace working certificates on an existing {productname} installation with a new third party certificate.
+In this case, you can replace the old CA certificate RPM with the new one, and update the database.
+
+
+
+.Procedure: Replacing Existing Certificates
```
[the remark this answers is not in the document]
Loquacity: I'm not sure, I really struggled to work this procedure out, honestly. It seems significantly different from the previous ones (and long, too!)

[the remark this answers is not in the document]
mcalmer: I agree, that the procedure before is already very long. I will try to do a PR on top of this one and enhance this one with the missing pieces. After this we can have a look and decide what we want to do.

Loquacity:
> I think there is a general issue with word usage. I think here we use "third party certs" and "custom certs" as equivalents. If this is correct we should settle on one term. In the intro we could say that custom certs are all certs not shipped with suma (self-created ones and foreign/third party ones).
> 
> I hope my comment makes sense.

I was interpreting custom certs as self-created ones, and third-party as coming from a certificate authority. But yes, I think it's confusing, we need to be really clear.

mcalmer:
> I was interpreting custom certs as self-created ones, and third-party as coming from a certificate authority. But yes, I think it's confusing, we need to be really clear.

I think we need to be clear between to cases. 

1. Automatically generated certs using rhn-ssl-tool during setup or use this tool later without --rpm-only option
2. Somebody else generated the certificates and we import them only. This is the new chapter.

modules/administration/pages/ssl-certs.adoc
```diff
@@ -1,9 +1,17 @@
...
-You can use custom certificates with {productname} and {productname} Proxy.
+{productname} uses SSL certificates to ensure that clients are registered to the correct server.
 
-This section covers how to use a third party SSL certificate authority with a new {productname} installation, and replacing existing certificate with new custom certificates.
+When you create an SSL certificate for your server, the certificate includes the host name of the server it is associated with.
+The certificate is then deployed on the server, and when clients attempt to connect, they check the certificate details.
+If the certificate contains the name of the server they are attempting to connect to, the connection will be successful.
+The client will also check if the certificate has been signed by a trusted certificate authority.
+
+By default, {productname} uses a self-signed certificate.
```
mcalmer: Found this while google for `self-signed certificate`:

`In cryptography and computer security, a self-signed certificate is a certificate that is not signed by a certificate authority (CA). `

This means SUSE Manager is **not** using this. We generate a CA and sign our server certificate with it.

Loquacity: But we're not a certificate authority ... ?

mcalmer: Seems I understood self-signed different to what people outside think it is. Maybe you can verify with wikipedia, etc. if there is a common understanding out there.

For me, self-signed is a certificate which is signed by itself and is at the same time CA and server certificate. This is something we cannot use at SUSE Manager.

But it seems that people also understand self-signed, as signed by an own PKI. You created your CA and use it to sign your own certificates with it. This is ok, and this is what we also do in the SUSE Manager setup procedure.
So here you are your own CA. 

So we need to differ between "internally" and "externally" generated certificates.

modules/administration/pages/ssl-certs.adoc
```diff
@@ -294,3 +310,47 @@ To complete the process, you need to update the database with this command:
...
+This file is not required, but you can provide it, by copying it into the server directory as `server.csr`.
++
+----
+cp <Certificate_Request_File>.csr /root/ssl-build/<Server_Name>/server.csr
+----
++
+When you repeat the command to generate the server certificate RPM, the warning should not be shown anymore.
+
+Now you have an ssl-build directory similar to the one generated by "Procedure: Creating New Certificates" from the chapter before.
+You can now continue with the "Procedure: Create Combined Certificate RPMs" from the "Create and Replace CA and Server Certificates" chapter.
```
mcalmer: @Loquacity @keichwa you are the asciidoc experts. I think there should be links to the sections. But I have no idea how to do it.

modules/administration/pages/ssl-certs.adoc
Loquacity: OK, let me figure this out.

Loquacity: I suspect the problem stems from definitions of a CA, not from definitions of a self-signed certificate. If you define a CA as any organisation that signs certificates, it's a bit murky. If a CA is an organisation that will perform research and verify a certificate for *other* organisations, then the difference becomes clear. A self-signed certificate is a certificate signed by the organisation that intends to use it (one website I looked at suggested that it was the same as approving your own passport application, which is oversimplified, but a good visual), and not by an external CA. Either way, the certificate must be signed, it's just a question of who does the signing, and under what authority.

Loquacity: So now I would like to know what we mean when we say "custom" certificate, because the way we've used the term in this doc makes me think it's neither a self-signed cert, or one verified by a CA.

Loquacity: And once I have the various certificate types clear in my head, I'm going to split this chapter, I think.

mcalmer: I think we need to distinguish between certificates generated by our tools or getting files from external. That external can be an own PKI from the infra team or a third party CA. But this does not matter in our case. 

We either generate the certs with out tools or we get the certificates and keys from somewhere else.

Now the question is: how to name this properly

mcalmer: Maybe: self generated(or SUSE Manager generated) vs. imported ?

modules/administration/pages/ssl-certs-imported.adoc
```diff
@@ -0,0 +1,116 @@
...
+. Install the {productname} Proxy according to the instructions in xref:installation:install-intro.adoc[].
+. Complete the initial setup according to xref:installation:proxy-setup.adoc[].
+. At the command prompt, run:
++
+----
+configure-proxy.sh
+----
+. At the ``Do you want to import existing certificates?`` prompt, type kbd:[y].
+. Follow the prompts to complete setup.
+
```
mcalmer: A good place to add a Note that all server certificates for server and proxies must be singed by the same CA.

modules/administration/pages/ssl-certs-selfsigned.adoc
```diff
@@ -1,9 +1,16 @@
-[[custom-ssl]]
-= Custom SSL Certificates
+[[ssl-certs-selfsigned]]
+= Self-Signed SSL Certificates
 
-You can use custom certificates with {productname} and {productname} Proxy.
+By default, {productname} uses a self-signed certificate.
+In this case, the certificate is created and signed by {productname}.
+This method does not use a third party certificate authority to guarantee that the details of the certificate are correct.
```
mcalmer: This is not wong, but I would like to add that the and independent party check the details.

```suggestion
This method does not use an independent third party certificate authority to guarantee that the details of the certificate are correct.
```

The background: If you use it internally only, why do you need an "independent" authority to check the details?
I would say nobody is so trustworthy than yourself, isn't it?

modules/administration/pages/ssl-certs-selfsigned.adoc
```diff
@@ -1,9 +1,16 @@
...
+
+It provides basic checking between the {productname} Server and the clients, but will not protect you against man-in-the-middle attacks.
```
mcalmer: This is wrong. Please remove this sentence. If you keep the private key secret, it provide the same protection as a third party certificate.

modules/administration/pages/ssl-certs-selfsigned.adoc
```diff
@@ -17,63 +24,13 @@ The host name of the SSL keys and certificates must match the fully qualified ho
 You can set the host names in the ``X509v3 Subject Alternative Name`` section of the certificate.
 You can also list multiple host names if your environment requires it.
```
mcalmer: This apply only to Import Certificate. Here we generate everything on our own. This can be removed

modules/administration/pages/ssl-certs-selfsigned.adoc
mcalmer: (only line 8)

modules/administration/pages/ssl-certs-selfsigned.adoc
Loquacity: This is mentioned in the intro, but I'll copy it across.

PR #856, opened 2019-10-29
more xrefs

keichwa:
for activation keys and product/channel selection.


Loquacity:
image

@jcayouette this is starting to bug me now: because we're only testing master, not the branch, we don't find out about errors until after the PR has merged, and then every single test fails until we merge a fix. We need to be finding the errors *before* merge.

keichwa:

> We need to be finding the errors *before* merge.

+1

And it would be good if it would not complain about submissions that are
not related with the faulty commit.

-- 
                           90409 Nürnberg, Germany

Loquacity:
> We need to be finding the errors *before* merge.
> +1 And it would be good if it would not complain about submissions that are not related with the faulty commit.

If I'm not mistaken, then changing the CI to run only on the feature branch should fix that. In the mean time, we might end up with a conflict, because the broken xref has been fixed here, as well as in https://github.com/SUSE/doc-susemanager/pull/860

PR #857, opened 2019-10-29
Add the instruction to disable the fqdns grain on the registerd systems(bsc#1134860)

admd:
Add the instruction to disable the fqdns grain on the registered systems

### DIFF
--- a/modules/client-configuration/pages/tshoot-clients.adoc
+++ b/modules/client-configuration/pages/tshoot-clients.adoc
@@ -79,3 +79,31 @@ SSLProtocol all -SSLv2 -SSLv3
 
 This will need to be done manually on the server, and with a Salt state on the Proxy.
 Restart the [systemitem]``apache`` service on each system after making the changes.
+
+
+
+== Disabling the FQDNS grain
+
+FQDNS grain returns the list of all the FQDN of the system. Ideally it should be pretty fast to collect this list but if the DNS settings on the system has some problem or misconfigured for some reason then this grain's
+calculation could get pretty expensive and as during name resolution Python's GIL is acquired so the salt-minion process could completely get stuck.
+
+To guard from this problem, a new flag in salt has been introduced which will disable this grain. Additionally salt network module has been extended to have fqdns function instead which returns the same result as FQDNS grain
+but without the problem described above.
+
+
+Newly registered Salt clients will have the FQDNS grain disabled by default.
+
+Disable the FQDNS grain on the existing registered salt clients by running this command on the server:
+----
+salt '*' state.sls util.mgr_disable_fqdns_grain
+----
+
+This will restart the clients and generate some Salt events to be processed by the server.
+If you have a large number of clients, handling these events could create excessive load.
+To avoid this, you can execute the command in batch mode with this command:
+----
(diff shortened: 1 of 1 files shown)


admd:
> Not sure whether this is the right chapter. To me it's not really related to rate limiting, but generally notable for performance and in particular in high scale cases.

Yeah, I wasn't sure either and couldn't find more suitable chapter. The only reason I put it there because it was kind of close to another kind of similar scenario. I can, of course, move it to a more appropriate chapter if Doc team could point me to.

Loquacity:
> > Not sure whether this is the right chapter. To me it's not really related to rate limiting, but generally notable for performance and in particular in high scale cases.
> 
> Yeah, I wasn't sure either and couldn't find more suitable chapter. The only reason I put it there because it was kind of close to another kind of similar scenario. I can, of course, move it to a more appropriate chapter if Doc team could point me to.

It looks like troubleshooting to me ...

moio:
@Loquacity this is not really troubleshooting, it's a best practice for all users that onboarded any minion before the latest version. It can be seen as an optimization, and will be cited in the release notes.

It only affects minions, in principle independently whether they are connected to an Uyuni/SUSE Manager Server or not. In practice, this is especially useful if they are.

Loquacity:
> @Loquacity this is not really troubleshooting, it's a best practice for all users that onboarded any minion before the latest version. It can be seen as an optimization, and will be cited in the release notes.
> 
> It only affects minions, in principle independently whether they are connected to an Uyuni/SUSE Manager Server or not. In practice, this is especially useful if they are.

So I think we're agreeing on large scale/tuning then?

moio:
> So I think we're agreeing on large scale/tuning then?

FMPOV, only if we do not have a better place, as this is useful regardless of scale.

Loquacity:
> > So I think we're agreeing on large scale/tuning then?
> 
> FMPOV, only if we do not have a better place, as this is useful regardless of scale.

So what about the client config guide then?

moio:
Sounds much better!

Loquacity:
@admd Do you need some help with this one?

admd:
Hi @Loquacity, I am so sorry. I was of the view that doc-squad will take care of moving this text to the correct part and totally forgot about this, my fault. Can you please guide me to the page in the client config guide where I should add this text?

Loquacity:
> Hi @Loquacity, I am so sorry. I was of the view that doc-squad will take care of moving this text to the correct part and totally forgot about this, my fault. 

No problem, I'm happy to take this on if that's what you would prefer.

> Can you please guide me to the page in the client config guide where I should add this text?

If you don't think it belongs in troubleshooting, then I would make it into its own section, which can go into the nav around about the same place as SSM stuff.

Let me know if you want a hand.

admd:
Hi @Loquacity , I have moved the text to the troubleshooting page in client config.

Loquacity:
> Hi @Loquacity , I have moved the text to the troubleshooting page in client config.

In this repo, or in the Uyuni repo?

Loquacity:
Please review https://github.com/uyuni-project/uyuni-docs/pull/28

PR #861, opened 2019-10-30
Update ssl-certs.adoc

keichwa:
the wrong `susemanager_setup` spelling is creeping in again and again.
by default, we no longer want to enforce the graphical yast; use `yast` instead of `yast2`.
later on, we want to reference this label.


modules/administration/pages/ssl-certs.adoc
```diff
@@ -290,7 +291,7 @@ rhnpush -c ssl-ca-channel --nosig \
 
 
 When you have the new certificate in the channel, you can use the {productname} {webui} to update it on all clients and proxies, by synchronizing them with the channel.
-Alternatively, for Salt clients, you can use menu:Salt[Remote Commands], or apply the highstate.
+Alternatively, for Salt Clients, you can use menu:Salt[Remote Commands], or apply the highstate.
```
Loquacity:
```suggestion
Alternatively, for Salt clients, you can use menu:Salt[Remote Commands], or apply the highstate.
```

PR #875, opened 2019-11-06
web login timeout

Loquacity:
SUSE/spacewalk#9901

### DIFF
--- a/modules/administration/pages/tshoot-logintimeout.adoc
+++ b/modules/administration/pages/tshoot-logintimeout.adoc
@@ -1,0 +1,55 @@
+[[troubleshooting-login-timeout]]
+= Troubleshooting Login Timeouts
+
+////
+PUT THIS COMMENT AT THE TOP OF TROUBLESHOOTING SECTIONS
+
+Troubleshooting format:
+
+One sentence each:
+Cause: What created the problem?
+Consequence: What does the user see when this happens?
+Fix: What can the user do to fix this problem?
+Result: What happens after the user has completed the fix?
+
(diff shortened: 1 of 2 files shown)


modules/administration/pages/tshoot-logintimeout.adoc
```diff
@@ -0,0 +1,55 @@
...
+Result: What happens after the user has completed the fix?
+
+If more detailed instructions are required, put them in a "Resolving" procedure:
+.Procedure: Resolving Widget Wobbles
+. First step
+. Another step
+. Last step
+////
+
+By default, the {productname} {webui} will require users to log in again after thirty minutes.
```
keichwa:
```suggestion
By default, the {productname} {webui} will require users to log in again after 30{nbsp}minutes.
```

modules/administration/pages/tshoot-logintimeout.adoc
```diff
@@ -0,0 +1,55 @@
...
+.Procedure: Adjusting the {webui} Login Timeout Value
+. Stop services:
++
+----
+spacewalk-service stop
+----
+. Open [path]``/etc/rhn/rhn.conf`` and add or edit this line to include the new timeout value in seconds:
++
+----
+web.session_database_lifetime = <Timeout_Value_in_Seconds>
```
mbologna: nit: do we have the same way to specify a parameter to be filled between <angled brackets> in other parts of the documentation?

Loquacity: I don't think we have a good solution to replaceable parameters at all, honestly.

PR #883, opened 2019-11-07
Ref autoyast #9892 ke

keichwa:
close https://github.com/SUSE/spacewalk/issues/9892

We now can probably delete autoyast.adoc in client-config; we now have a newer version as client-configuration/pages/client-automating-installation.adoc.
Only creating and using an ISO boot image is missing; I think info about this the user can find in the autoyast documentation, if needed.


keichwa:
thanks, @mcalmer

PR #889, opened 2019-11-13
Remove port 8050

cbosdo:
Since uyuni-project/uyuni#1579 and SUSE/spacewalk#10044 the client doesn't need access to port 8050 on the server.


cbosdo:
Ooops... looks like I haven't pushed to the right thing

PR #900, opened 2019-11-26
Cobbler #7496 ke

keichwa:
see https://github.com/SUSE/spacewalk/issues/7496

### DIFF
--- a/modules/client-configuration/pages/cobbler.adoc
+++ b/modules/client-configuration/pages/cobbler.adoc
@@ -23,36 +23,11 @@ For more information about Cobbler, see https://cobbler.readthedocs.io.
 
 
 
-////
-=== Background Information about the Synchronization Process
-
-// Taken from https://bugzilla.suse.com/show_bug.cgi?id=1141603
-// To be edited!!!
-
-A sync is basically a rebuild of every file cobbler touched.
-
-1. Run Pre-Sync-Triggers (this can be any number of shell scripts to my knowledge)
-2. Delete all files and directories which are not allowed in /var/www/cobbler
-3. Create all needed folders
-4. Delete all elements inside the folders.
-5. Generate the tftpd directory
-6. Write the DHCP files (if management is enabled)
-7. Do the same with DNS
-8. Clean the Cache
-9. If rsync management is enabled do that
-10. Run Post-Sync-Triggers (this can also be any number of shell scripts to my knowledge)
-
-////
-
-
-
 == Cobbler Requirements
 
 To use Cobbler for system installation with PXE, you will require a TFTP server.
 {productname} installs a TFTP server by default.
 To PXE boot systems, you will require a DHCP server, or have access to a network DHCP server.
-
-
 
 [IMPORTANT]
 ====
@@ -206,6 +181,32 @@ This will fail if you have not configured the proxies correctly.
 ====
 If you reinstall an already configured proxy and want to push all the files again, you must remove the cache file at [path]``/var/lib/cobbler/pxe_cache.json`` before you call [command]``cobbler sync``.
 ====
+
+=== [command]``cobbler sync``
+
+////
+=== Background Information about the Synchronization Process
+
+// Taken from https://bugzilla.suse.com/show_bug.cgi?id=1141603
+// To be edited!!!
+
+A sync is basically a rebuild of every file cobbler touched.
+
+1. Run Pre-Sync-Triggers (this can be any number of shell scripts to my knowledge)
+2. Delete all files and directories which are not allowed in /var/www/cobbler
+3. Create all needed folders
+4. Delete all elements inside the folders.
+5. Generate the tftpd directory
+6. Write the DHCP files (if management is enabled)
+7. Do the same with DNS
+8. Clean the Cache
+9. If rsync management is enabled do that
+10. Run Post-Sync-Triggers (this can also be any number of shell scripts to my knowledge)
+
+////
+
+On {productname}, [command]``cobbler sync`` does the following actions:
+
 
 
 
--- a/modules/client-configuration/pages/cobbler.adoc
+++ b/modules/client-configuration/pages/cobbler.adoc
@@ -2,7 +2,6 @@
 = Cobbler
 
 Cobbler is an installation server that allows you to perform unattended system installations.
-It can be used on server, client, or guest systems, and in virtual environments.
 Cobbler is installed on the {productname} Server.
 
 [IMPORTANT]
--- a/modules/client-configuration/pages/cobbler.adoc
+++ b/modules/client-configuration/pages/cobbler.adoc
@@ -13,7 +13,6 @@ Only supported features are documented here.
 This section explains the Cobbler features most commonly used with {productname}:
 
 * Installation environment analysis using the [command]``cobbler check`` command
-* Multi-site installation server configuration using the [command]``cobbler replicate`` command
 * Virtual machine guest installation automation with the [command]``koan`` client-side tool
 * Building installation ISOs with PXE-like menus using the [command]``cobbler buildiso`` command (for {productname} systems with x86_64 architecture)
 
--- a/modules/client-configuration/pages/cobbler.adoc
+++ b/modules/client-configuration/pages/cobbler.adoc
@@ -3,6 +3,13 @@
 
 Cobbler is an installation server that allows you to perform unattended system installations.
 It can be used on server, client, or guest systems, and in virtual environments.
+On {productname} Cobbler is installed on the server.
+
+[IMPORTANT]
+====
+{suse} only supports Cobbler functions that are available in the {productname} {webui}, or through the {productname} API.
+All features documented here are supported.
+====
 
 This section explains the Cobbler features most commonly used with {productname}:
 
@@ -13,11 +20,6 @@ This section explains the Cobbler features most commonly used with {productname}
 
 For more detailed Cobbler documentation, see http://cobbler.github.io/manuals/.
 
-[IMPORTANT]
-====
-{suse} only supports Cobbler functions that are available in the {productname} {webui}, or through the {productname} API.
-All features documented here are supported.
-====
 
 [IMPORTANT]
 ====
@@ -31,7 +33,7 @@ If you intend to use your installation with {smr} formulas, do not configure Cob
 // Taken from https://bugzilla.suse.com/show_bug.cgi?id=1141603
 // To be edited!!!
 
-A sync is basically a rebuild of every file cobbler touched. I will try give you an impression with an numbered list.
+A sync is basically a rebuild of every file cobbler touched.
 
 1. Run Pre-Sync-Triggers (this can be any number of shell scripts to my knowledge)
 2. Delete all files and directories which are not allowed in /var/www/cobbler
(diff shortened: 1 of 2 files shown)


modules/client-configuration/pages/cobbler.adoc
```diff
@@ -53,9 +55,10 @@ A sync is basically a rebuild of every file cobbler touched. I will try give you
 To use Cobbler for system installation with PXE, you will require a TFTP server.
 {productname} installs a TFTP server by default.
 To PXE boot systems, you will require a DHCP server, or have access to a network DHCP server.
+// ke, 2019-11-26: Maybe, we should say that this on SUMA has to be the address of the SUMA server?
 Edit the [path]``/etc/dhcp.conf`` configuration file to change [option]``next-server`` to the hostname or IP address of your Cobbler server.
 
-Cobbler requires an open HTTP port to synchronize data between the Server and the Proxy.
+Cobbler requires an open HTTP port to synchronize data between the server and the proxy.
```
mcalmer: Question: Do we explain tftpsync somewhere else ? Because cobbler itself does not require these ports to be open. At least not for the features of cobbler we use.

But we have such a sync mechanism implemented as a trigger of cobbler. This is tftpsync. You need to explicitly configure it. So if you do this, you need to open the ports. But by default the ports are not needed.

On the other hand: "HTTP(s)" is used for all the other features of SUSE Manager Server and Proxy. Closing them, completly makes the product unusable.

But in this context here we can remove it if we have a chapter about tftpsync.

mcalmer: Just found it here in the same page. 

```
You can also synchronize Cobbler-generated TFTP contents to a SUSE Manager Proxy.
...
```
Here we can add the note, that the sync happens via https and the ports needs to be open.

keichwa: ok, I moved this info.  thanks!

keichwa:
@SchoolGuy I'd appreciate if you would have the time to check this PR.  Then it would be great if you'd read the complete  `modules/client-configuration/pages/cobbler.adoc` file and edit/remove everything that's not supported on SUMA.

Thanks for your help!

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -3,6 +3,13 @@
 
 Cobbler is an installation server that allows you to perform unattended system installations.
 It can be used on server, client, or guest systems, and in virtual environments.
+On {productname} Cobbler is installed on the server.
```
Loquacity:
```suggestion
Cobbler is installed on the {productname} Server.
```

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -3,6 +3,13 @@
...
+
+[IMPORTANT]
+====
+{suse} only supports Cobbler functions that are available in the {productname} {webui}, or through the {productname} API.
+All features documented here are supported.
```
Loquacity: I wonder if it works better to turn this around:

```suggestion
Only supported features are documented here.
```

The most important info is presumably the supported features, rather than the documented features?

modules/client-configuration/pages/cobbler.adoc
keichwa: Sounds better, but then we'd better remove the sentence before to make the message clear.

modules/client-configuration/pages/cobbler.adoc
keichwa: yes, much better, IMO.

mcalmer:
> `Multi-site installation server configuration using the [command]``cobbler replicate`` command`

This is not supported with SUSE Manager. We have tftpsync which is different.

mcalmer:
> [IMPORTANT]
> ====
> Cobbler is not currently supported within {smr} environments.
> If you intend to use your installation with {smr} formulas, do not configure Cobbler.
> ====

I do not understand this. Cobbler is always automatically configured and you cannot run SUSE Manager without it. What does it has to do with "formulas"? Can we find out what we want to say with this note?

mcalmer:
> This section explains the Cobbler features most commonly used with {productname}:

If we explain some commands and functionality we should add `cobbler sync`. This command is triggered from SUSE Manager Server and generate the tftp boot environment on the server.

This is the main functionality of cobbler in combination with SUSE Manager.

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -53,12 +54,9 @@ A sync is basically a rebuild of every file cobbler touched. I will try give you
 To use Cobbler for system installation with PXE, you will require a TFTP server.
 {productname} installs a TFTP server by default.
 To PXE boot systems, you will require a DHCP server, or have access to a network DHCP server.
+// ke, 2019-11-26: Maybe, we should say that this on SUMA has to be the address of the SUMA server?
 Edit the [path]``/etc/dhcp.conf`` configuration file to change [option]``next-server`` to the hostname or IP address of your Cobbler server.
```
mcalmer: The DHCP Server is not a feature of cobbler nor it is provided by SUSE Manager. So we talk about any dhcp server in the company network. It can be dhcpd from a SLES or RedHat, but it can also be dnsmasq or a Microsoft DHCP server.

Describing "how to configure it" might be difficult. We can say "If your dhcp server is a XXXXX you configure it this way" but if you use something else, you need to "adapt" this description.

cobbler has the feature to use its template mechanism to configure dhcp, but this is **not supported** with SUSE Manager.

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -200,13 +205,13 @@ configure-tftpsync.sh proxy1.example.com proxy2.example.com
...
 [NOTE]
 ====
 If you reinstall an already configured proxy and want to push all the files again, you must remove the cache file at [path]``/var/lib/cobbler/pxe_cache.json`` before you call [command]``cobbler sync``.
 ====
 
 
+
 == Synchronize and Start the Cobbler Service
 
 The {productname} Server must be able to access the {productname} Proxy systems directly.
```
mcalmer:
```suggestion
When tftpsync is configured, the {productname} Server must be able to access the {productname} Proxy systems directly.
```

mcalmer: I do not understand the next sentense:

> You cannot push using a proxy.

mcalmer:
> Kickstart Templates

This explains `variables` and `snippets` with Kickstart Profiles. This is ok, but we should say somewhere prominent, that `variables` and `snippets` can also be used in "Autoyast Profiles".

mcalmer:
> Note: Building ISOs with Cobbler is not supported on {ibmz}.

I think also not on ppc64le ? But I am not sure if this can work or not.

mcalmer:
> Enable Bare Metal Systems Management
> ...
> The process can take a few minutes, and the system will automatically shut down once it is complete. After the reboot, the system will appear in the Systems list.

There is no "reboot". I would say:

```
The process can take a few minutes, and the system will automatically shut down once it is complete. The system should now be visible in the Systems List.
```

SchoolGuy:
Also I would finally remove the commented out parts of the document. There are no plans or efforts to my knowledge to support them in the future. @keichwa

keichwa:
@mcalmer and @SchoolGuy , thanks for all your feedback.  I'll work on this now.

keichwa:
> > `` Multi-site installation server configuration using the [command]``cobbler replicate`` command ``
> 
> This is not supported with SUSE Manager. We have tftpsync which is different.
I'll remove this item (later on, it was not explained, though ;)

keichwa:
```
[IMPORTANT]
Cobbler is not currently supported within {smr} environments.
If you intend to use your installation with {smr} formulas, do not configure Cobbler.
```

> I do not understand this. Cobbler is always automatically configured and you cannot run SUSE Manager without it. What does it has to do with "formulas"? Can we find out what we want to say with this note?

@aaannz , do you know why we have this admonition here?  Maybe, we must re-phrase it somehow.

aaannz:
Sure, this is to make sure that user will not run/enable cobbler managed tftp server. This limitation is currently only for branch proxy, but work is ongoing to enable terminal bootstrap directly from SUMA. Possibly in the 4.1 there will be this limitation even for SUMA for Retail or we'll need to figure out how can SMR and Cobbler run together - including other services like cobbler dhcp, etc.

keichwa:
> EDIT: Ok, so the cobbler and dhcp is not an issue. So only tftp for pxe boot environment is.

Does this mean that we can delete the admonition about cobbler and SMR here completely and better a different warning about SMR and  tftp for pxe boot somewhere else? Or shall we rephrase the existing warning here?

(BTW, @mcalmer asked initially.)

modules/client-configuration/pages/cobbler.adoc
keichwa: Ok, then I'd better just remove line 58 (and 57).

modules/client-configuration/pages/cobbler.adoc
keichwa: I'll delete this sentence.

keichwa:
> > Note: Building ISOs with Cobbler is not supported on {ibmz}.
> 
> I think also not on ppc64le ? But I am not sure if this can work or not.

@paususe I need your help please.  Do we have to mention such a limitation here (assuming that it exists).  Or is this something to add to the feature matrix?

mcalmer:
> Sure, this is to make sure that user will not run/enable cobbler managed tftp server. This limitation is currently only for branch proxy, but work is ongoing to enable terminal bootstrap directly from SUMA. Possibly in the 4.1 there will be this limitation even for SUMA for Retail or we'll need to figure out how can SMR and Cobbler run together ~- including other services like cobbler dhcp, etc~.
> 
> EDIT: Ok, so the cobbler and dhcp is not an issue. So only tftp for pxe boot environment is.

Currently it affect only Branch Server == Proxy. On a proxy we do not have cobbler, but you could do tftpsync. I think this is the part which would break things. (in 4.0). So we may write that you should not enable tftpsync to a Retail Branch Server. 

@aaannz when you introduce limits on the server in 4.1 it will become a bit difficult. Cobbler is used via API and you cannot just disable it. It will crash the UI at places you never thought about. So cobbler will run and will create /srv/tftpboot/ . Reserve enough time to do the research how to make it work.

keichwa:
I think a addressed all concerns.  I just want to add more details on `cobbler sync`.

SchoolGuy:
@keichwa How detailed do you need the docs? I will accordingly create upstream docs and then you can take from there what you need!

keichwa:
"Enno G." <notifications@github.com> writes:

> @keichwa How detailed do you need the docs?

What we now have is probably detailed enough.  I'm sure Lana will help
with faceliftig when we stop adding fixes.

The only thing that's missing is a description what exactly happens when
the user runs "cobbler sync" on a SUMA default installation.  If you
could provide this it would help a lot.

Some weeks ago, you provided this listing:

```
A sync is basically a rebuild of every file cobbler touched.

1. Run Pre-Sync-Triggers (this can be any number of shell scripts to my knowledge)
2. Delete all files and directories which are not allowed in /var/www/cobbler
3. Create all needed folders
4. Delete all elements inside the folders.
5. Generate the tftpd directory
6. Write the DHCP files (if management is enabled)
7. Do the same with DNS
8. Clean the Cache
9. If rsync management is enabled do that
10. Run Post-Sync-Triggers (this can also be any number of shell scripts to my knowledge)
```

I'm especially wondering whether all these steps are valid in our context?


-- 
                           90409 Nürnberg, Germany

SchoolGuy:
@keichwa Yes they are. SUMA does just nothing more then triggering a cobbler sync.

Just to mention it: There is also a lite sync. This sync works similar but only touches files and directories which are related to the change which triggered it. SUMA also adds/removes and edits systems which are in cobbler and thus triggers such lite-syncs.

keichwa:
Now I please want some final approvals.  Then I'll push it to the uyuni-docs repo (assuming that that's the thing to do).

keichwa:
closing here.  we now have the same PR (squashed) in uyuni-project.

PR #908, opened 2019-11-29
Docs: update unsupported CaaSP operations

mbologna:
As part of the merged RFC for forbidden CaaSP operations, this PR defines the operations that should not be done on a CaaSP node registered to Uyuni/SUSE Manager.
Ref: https://github.com/uyuni-project/uyuni-rfc/pull/26

### DIFF
--- a/modules/client-configuration/pages/vhm-caasp.adoc
+++ b/modules/client-configuration/pages/vhm-caasp.adoc
@@ -31,16 +31,16 @@ For more information, see https://documentation.suse.com/suse-caasp/4/html/caasp
 
 [WARNING]
 ====
-Do not use Salt, {productname} (either via UI or API) to perform these actions on any {caasp} nodes:
+When using Salt or {productname} (either via UI or API) on any {caasp} nodes:
 
-* Apply a patch (if the patch is marked as interactive)
-* Mark a system to automatically install patches
-* Perform an SP migration
-* Reboot a node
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/vhm-caasp.adoc
```diff
@@ -26,18 +26,22 @@ When you have added the {caasp} nodes to {productname}, you will be able to:
...
+* Mark a system to automatically install patches
+* Perform an SP migration
+* Reboot a node
+* Issue any power management action via Cobbler
+* Installing a package if it breaks or conflicts the `patterns-caasp-Node-x.y`
+* Removing a package if it breaks or conflicts or is one of the packages related with the `patterns-caasp-Node-x.y`
+* Upgrading a package if it breaks or conflicts or is one of the packages related with the `patterns-caasp-Node-x.y`
+
+{productname} will not stop you from issuing these operations.
+Issuing those operation may render your {caasp} cluster unusable.
```
Loquacity: I wonder if it would make more sense to put this in an admonition. I feel like listing a things of what not to do could be risky, as people tend to skim over stem sentences and jump straight to bullet points. I'm not sure we can trust that people will not read this as "oh, yeah, I can reboot nodes!"

Loquacity: Yep, that works :)

PR #909, opened 2019-12-11
Update clients-ubuntu.adoc

mantel:
sudo root access for onboarding

### DIFF
--- a/modules/client-configuration/pages/clients-ubuntu.adoc
+++ b/modules/client-configuration/pages/clients-ubuntu.adoc
@@ -238,7 +238,7 @@ Grant [command]``sudo`` access to the user by adding this line to the [filename]
 <user>  ALL=NOPASSWD: /usr/bin/python, /usr/bin/python2, /usr/bin/python3
 ----
 
-
+This passwordless root access is only needed for onboarding the client. Once the salt-minion is installed, this is no longer needed, because salt-minion is running with root privileges. So it is recommended to remove this line after the minion has been successfully onboarded.
 
 == Register Clients
 


keichwa:
Thanks Hubert!  I think we also need Lana who recently worked on this file.

modules/client-configuration/pages/clients-ubuntu.adoc
```diff
@@ -238,7 +238,7 @@ Grant [command]``sudo`` access to the user by adding this line to the [filename]
 <user>  ALL=NOPASSWD: /usr/bin/python, /usr/bin/python2, /usr/bin/python3
 ----
 
-
+This passwordless root access is only needed for onboarding the client. Once the salt-minion is installed, this is no longer needed, because salt-minion is running with root privileges. So it is recommended to remove this line after the minion has been successfully onboarded.
```
Loquacity:
```suggestion
This procedure grants root access without requiring a password, which is required for registering the client.
When the client is successfully installed it will run with root privileges, so the access is no longer required.
We recommend that you remove the line from the `sudoers` file after the client has been successfully installed.
```

Probably also a good idea to put this in an admonition (note?).

mantel:
Chances are high that we will not need this manual change at all in the future! I could verify that this permanent change to the sudoers configuration is not even needed when using other contact methods such as ssh-push. Even in these cases permanent root access is not needed.

Apart from that: Instead of modifying the main sudo file (/etc/sudoers), I would prefer to just drop in an include file with our stuff into /etc/sudoers.d

https://github.com/SUSE/spacewalk/issues/10246

Loquacity:
**DO NOT MERGE**

This needs to go into the Uyuni repo, not this one!

Loquacity:
Please review https://github.com/uyuni-project/uyuni-docs/pull/29

